PR #13, opened 2020-01-05
Add the ability to specify separate rtmp & media (B & O) hosts

mk-livepeer:
Currently, stream-tester assumes that the broadcaster and orchestrator nodes are hosted on the same machine.  It takes a "host" string argument along with "rtmp" & "media" integer arguments that designate the ports to be used on the single host.

This adds a new field for the media / orchestrator host name to be used in addition to the existing field that is used for the rtmp / broadcaster host name.

This was done in a fully-backward compatible way - If the new field is not provided, the application will continue to function with its original behavior.

blocks https://github.com/livepeer/livepeer-infra/issues/32
closes https://github.com/livepeer/stream-tester/issues/14

### DIFF
--- a/cmd/streamtester/streamtester.go
+++ b/cmd/streamtester/streamtester.go
@@ -30,7 +30,8 @@ func main() {
 	sim := flag.Uint("sim", 1, "Number of simulteneous streams to stream")
 	repeat := flag.Uint("repeat", 1, "Number of time to repeat")
 	profiles := flag.Int("profiles", 2, "Number of transcoding profiles configured on broadcaster")
-	host := flag.String("host", "localhost", "Broadcaster's host name")
+	bhost := flag.String("host", "localhost", "Broadcaster's host name")
+	ohost := flag.String("orch", "", "Orchestrator's host name - Broadcaster host will be used by default if not specified")
 	rtmp := flag.String("rtmp", "1935", "RTMP port number")
 	media := flag.String("media", "8935", "Media port number")
 	stime := flag.String("time", "", "Time to stream streams (40s, 4m, 24h45m). Not compatible with repeat option.")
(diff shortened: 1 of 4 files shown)


darkdarkdragon:
@mk-livepeer I've merged, but after that I've renamed some parameters.

j0sh:
Can someone clarify a bit more what the overall goals here are? My interpretation is that this PR :

* Adds a new flag indicating the orchestrator host
* Orchestrator host is used to download segments

However, the broadcaster could be using any number of orchestrators. This patch would only work if the B was hard-coded to a single O - which is not really recommended practice. Is my understanding correct? Moreover, how does this help unblock https://github.com/livepeer/livepeer-infra/issues/32 which doesn't really make any mention of downloading from the O directly?

mkrufky:
This was a necessary change in order to support kubernetes and docker-compose, for example:
```diff
--- a/bot-monitoring/docker-compose.yml
+++ b/bot-monitoring/docker-compose.yml
@@ -1,5 +1,11 @@
 version: '3.5'
 services:
+  streamtester:
+    depends_on:
+      - broadcaster
+      - orchestrator
+    image: livepeer/streamtester
+    command: './streamtester -host broadcaster -media-host orchestrator'
   orchestrator:
     image: livepeer/go-livepeer:master
     command: '-orchestrator --network rinkeby -orchSecret /root/osecret.txt -serviceAddr orchestrator:8935 -initializeRound true -ethPassword /root/pw.txt -cliAddr :7935 -monitor=true'
```

The two nodes will never have the same hostname or IP in this scenario, nor within K8S.  Adding an optional command line argument is the correct solution.

darkdarkdragon:
@mkrufky 
> The two nodes will never have the same hostname or IP in this scenario, nor within K8S. Adding an optional command line argument is the necessary solution

Still, I don't understand what you're trying to do. You're streaming into B node, why do you want to read data back from O node?

mkrufky:
My mistake.  I thought we should be reading from the O, I didn't realize that the B also serves segments on 8935.  Should we revert this?

j0sh:
Reverting sounds good, since this feature may lead to more confusion if someone else tries to use it.

PR #59, opened 2020-11-10
change consul parameters

iameli:
Instead of hitting a service, we can create our own custom service record and forward traffic wherever we want. So this sets us up to route `/hls/{playbackId}` to `https://mdw-staging-mist-server-0.livepeer.monster`, which is ready for global replication on ingest URLs.

### DIFF
--- a/cmd/mist-api-connector/mist-api-connector.go
+++ b/cmd/mist-api-connector/mist-api-connector.go
@@ -36,7 +36,8 @@ func main() {
 	apiToken := fs.String("api-token", "", "Token of the Livepeer API to be used by the Mist server")
 	apiServer := fs.String("api-server", livepeer.ACServer, "Livepeer API server to use")
 	consulURI := fs.String("consul", "", "Base URL to access Consul (for example: http://localhost:8500)")
-	mistServiceNameInTraefik := fs.String("mist-service-name", "", "Name of the service for this Mist inside Traefik (to be put in Consul)")
+	playbackDomain := fs.String("playback-domain", "", "domain to create consul routes for (ex: playback.livepeer.live)")
+	mistURL := fs.String("mist-url", "", "external URL of this Mist instance (to be put in Consul) (ex: https://mist-server-0.livepeer.live)")
 	_ = fs.String("config", "", "config file (optional)")
 
 	ff.Parse(fs, os.Args[1:],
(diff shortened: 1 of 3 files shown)


iameli:
Maybe `-mist-url` isn't the best name if we already have `-mist-host` and `-mist-port`. Could call it `-external-mist-url` or `-consul-mist-url` or something.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -309,8 +312,14 @@ func (mc *mac) handleDefaultStreamTrigger(w http.ResponseWriter, r *http.Request
...
 		go func() {
-			err := consul.PutKeys(mc.consulURL, traefikKeyPathBase+streamKey+"/rule", fmt.Sprintf(traefikRuleTemplate, streamKey),
-				traefikKeyPathBase+streamKey+"/service", mc.mistServiceName)
+			err := consul.PutKeys(
+				mc.consulURL,
+				traefikKeyPathRouters+streamKey+"/rule",
+				fmt.Sprintf(traefikRuleTemplate, mc.playbackDomain, streamKey),
+				traefikKeyPathRouters+streamKey+"/service",
+				streamKey,
+				traefikKeyPathServices+streamKey+"/loadbalancer/servers/0/url",
```
darkdarkdragon, line 321:
```suggestion
				traefikKeyPathServices+streamKey+"/loadbalancer/servers/0/url",
				mc.mistURL,
```

cmd/mist-api-connector/mist-api-connector.go
```diff
@@ -36,7 +36,8 @@ func main() {
 	apiToken := fs.String("api-token", "", "Token of the Livepeer API to be used by the Mist server")
 	apiServer := fs.String("api-server", livepeer.ACServer, "Livepeer API server to use")
 	consulURI := fs.String("consul", "", "Base URL to access Consul (for example: http://localhost:8500)")
-	mistServiceNameInTraefik := fs.String("mist-service-name", "", "Name of the service for this Mist inside Traefik (to be put in Consul)")
+	playbackDomain := fs.String("playback-domain", "", "domain to create consul routes for (ex: playback.livepeer.live)")
+	mistURL := fs.String("mist-url", "", "external URL of this Mist instance (to be put in Consul) (ex: https://mist-server-0.livepeer.live)")
```
darkdarkdragon, line 40:
```suggestion
	mistURL := fs.String("consul-mist-url", "", "external URL of this Mist instance (to be put in Consul) (ex: https://mist-server-0.livepeer.live)")
```

darkdarkdragon (changes requested):
Small fix

darkdarkdragon:
> So this sets us up to route /hls/{playbackId} to https://mdw-staging-mist-server-0.livepeer.monster, which is ready for global replication on ingest URLs.

Interesting - I thought internal addresses should be used. But if it works than OK.

iameli:
Yeah, I tested it with `https://icanhazip.com` and it worked perfectly. Surprised me too but very useful. We can basically configure Traefik as an HTTPS proxy into anything we want now. cc @yondonfu, maybe useful for your T reverse-proxy work?

PR #68, opened 2020-12-18
orch-tester: random sample segments

kyriediculous:
If enabled using `-randomsample=true` orch-tester will randomly sample a segment and write it to the provided GCP bucket (`-gsbucket` and  `- gskey`) 

1. For each orchestrator pick a random video profile and seqNo 
2. Start the stream 
3. As the stream is started start fetching the target segment from the broadcaster
4. When the stream finishes stop fetching segments from the broadcaster
5. If no segment is found query the entire last known playlist for the stream's manifest ID
6. from that playlist randomly get a segment

### DIFF
--- a/cmd/orch-tester/orch_tester.go
+++ b/cmd/orch-tester/orch_tester.go
@@ -9,9 +9,11 @@ import (
 	"encoding/json"
 	"errors"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"math"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -20,6 +22,9 @@ import (
 
 	"github.com/golang/glog"
 	apiModels "github.com/livepeer/leaderboard-serverless/models"
+	"github.com/livepeer/m3u8"
+	"github.com/livepeer/stream-tester/internal/testers"
+	"github.com/livepeer/stream-tester/model"
 	streamerModel "github.com/livepeer/stream-tester/model"
 	promClient "github.com/prometheus/client_golang/api"
 	promAPI "github.com/prometheus/client_golang/api/prometheus/v1"
@@ -32,6 +37,10 @@ const bcastMediaPort = "8935"
 const bcastRTMPPort = "1935"
 const defaultBcast = "127.0.0.1"
 const httpTimeout = 8 * time.Second
+
+const numSegments = 15
+
+var start time.Time
 
 func main() {
 	flag.Set("logtostderr", "true")
@@ -47,9 +56,15 @@ func main() {
 	// Video config
 	videoFile := flag.String("video", "official_test_source_2s_keys_24pfs_30s.mp4", "video file to use, has to be present in stream-tester root")
 	numProfiles := flag.Int("profiles", 3, "number of video profiles to use on the broadcaster")
-	presets := flag.String("presets", "P240p30fps16x9,P360p30fps16x9,P720p30fps16x9", "video profile presets to use for HTTP ingest")
+	presets := flag.String("presets", "P240p30fps16x9,P360p30fps16x9", "video profile presets to use for HTTP ingest")
 	repeat := flag.Int("repeat", 1, "number of times to repeat the stream")
 	simultaneous := flag.Int("simultaneous", 1, "number of times to run the stream simultaneously")
+
+	// randomSample will sample one transcoded rendition and source segment randomly per stream
+	randomSample := flag.Bool("randomsample", false, "randomly sample a source and transcoded segment per stream")
+	gsBucket := flag.String("gsbucket", "", "Google storage bucket to store segments")
+	gsKey := flag.String("gskey", "", "Google Storage private key (in json format)")
+
 	flag.Parse()
 
 	if *region == "" {
@@ -76,6 +91,11 @@ func main() {
 		log.Fatal(err)
 	}
 
+	broadcasterURL, err := defaultAddr(fmt.Sprintf("%v:%v", *broadcaster, *media), "127.0.0.1", "8935")
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	rtmpUint, err := strconv.ParseUint(*rtmp, 10, 16)
 	if err != nil {
 		log.Fatal(err)
@@ -86,7 +106,9 @@ func main() {
 		log.Fatal(err)
 	}
 
-	streamer, err := newStreamerClient(streamTesterURL, metricsURL, leaderboardURL, *leaderboardSecret, subgraphURL)
+	profiles := strings.Split(*presets, ",")
+
+	streamer, err := newStreamerClient(streamTesterURL, metricsURL, leaderboardURL, *leaderboardSecret, subgraphURL, broadcasterURL, profiles)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -96,7 +118,12 @@ func main() {
 		log.Fatal(err)
 	}
 
-	refreshWait := 90 * time.Second
+	testers.Bucket = *gsBucket
+	testers.CredsJSON = *gsKey
+
+	refreshWait := 70 * time.Second
+
+	start = time.Now()
 
 	for _, o := range orchestrators {
 		time.Sleep(refreshWait)
@@ -123,7 +150,20 @@ func main() {
 			continue
 		}
 
-		glog.Infof("Started stream for orchestrator %v", o.Address)
+		var (
+			ctx    context.Context
+			cancel context.CancelFunc
+		)
+		randErr := make(chan error)
+		if *randomSample {
+			rand.Seed(time.Now().UnixNano())
+			ctx, cancel = context.WithCancel(context.Background())
+			go func(randErr chan error) {
+				randErr <- streamer.randomSample(ctx, mid, o.Address)
+			}(randErr)
+		}
+
+		glog.Infof("Started stream orchestrator=%v maninfestID=%v", o.Address, mid)
 
 		apiStats := &apiModels.Stats{
 			Region:       *region,
@@ -133,6 +173,7 @@ func main() {
 
 		// Make sure manifest ID exists before getting stats
 		time.Sleep(15 * time.Second)
+
 		// wait for stream to finish transcoding
 		streamerStats, err := streamer.getFinishedStats(mid)
 		if err != nil {
@@ -184,6 +225,15 @@ func main() {
 			glog.Error(err)
 			continue
 		}
+		// if we haven't found a random sample by now cancel and wait for the backup attempt to complete before returning
+		if *randomSample {
+			cancel()
+			err := <-randErr
+			if err != nil {
+				glog.Error(err)
+			}
+			continue
+		}
 	}
 }
 
@@ -218,11 +268,13 @@ type streamerClient struct {
 	leaderboardAddr   string
 	leaderboardSecret string
 	subgraph          string
+	broadcaster       string
 	metrics           promAPI.API
 	client            *http.Client
+	profiles          []string
 }
 
-func newStreamerClient(streamTesterURL, metricsURL, leaderboardAddr, leaderboardSecret, subgraph string) (*streamerClient, error) {
+func newStreamerClient(streamTesterURL, metricsURL, leaderboardAddr, leaderboardSecret, subgraph, broadcasterURL string, profiles []string) (*streamerClient, error) {
 	client, err := promClient.NewClient(promClient.Config{
 		Address: metricsURL,
 	})
@@ -239,6 +291,8 @@ func newStreamerClient(streamTesterURL, metricsURL, leaderboardAddr, leaderboard
 		leaderboardAddr:   leaderboardAddr,
 		leaderboardSecret: leaderboardSecret,
 		subgraph:          subgraph,
+		broadcaster:       broadcasterURL,
+		profiles:          profiles,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+				glog.Error(err)
+			}
+			break
+		case <-tick.C:
+			glog.V(model.DEBUG).Infof("Trying to download %v", renditionUrl)
+			rendition, err = s.downloadSegment(renditionUrl)
+			if err != nil {
+				glog.V(model.DEBUG).Info(err)
+				continue
+			}
```
yondonfu, line 632: Should there also be a debug statement for trying to download the source since we have one for the rendition?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+			source, err = s.downloadSegment(sourceUrl)
+			if err != nil {
+				glog.V(model.DEBUG).Info(err)
```
yondonfu, line 635: Should be Error instead of Info

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+			rendition, source, renditionUrl, err = s.sampleFromPlaylist(mid)
+			if err != nil {
+				glog.Error(err)
+			}
+			break
+		case <-tick.C:
+			glog.V(model.DEBUG).Infof("Trying to download %v", renditionUrl)
+			rendition, err = s.downloadSegment(renditionUrl)
+			if err != nil {
+				glog.V(model.DEBUG).Info(err)
```
yondonfu, line 630: Should be Error instead of Info

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+		}
+		break
+	}
+
+	if len(rendition) == 0 || len(source) == 0 {
+		return fmt.Errorf("no segments found")
+	}
+
+	y, m, d := start.Date()
+	baseFileName := fmt.Sprintf("%v-%v%v", y, m.String(), d)
```
yondonfu, line 648: Should the string format be YYYY-mm-dd instead of the current YYYY-mmdd?

yondonfu, line 648: Express month as an integer instead of using `.String()` to express the English full name for a more compact filename?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+		return err
+	}
+	glog.Infof("Wrote segment to storage file=%v", src)
+
+	rend, _, err := testers.SaveToExternalStorage(rendF, rendition)
+	if err != nil {
+		glog.Error(err)
+		return err
+	}
+	glog.Infof("Wrote segment to storage url=%v", rend)
```
yondonfu, line 665: "segment to storage url" -> "rendition to storage url" to clarify which file is being saved?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	u := strings.Split(renditionUrl, "/")
+	u = u[len(u)-2:]
+	rendF := fmt.Sprintf("%v/%v/%v-%v", baseFileName, orch, mid, strings.Join(u, "-"))
+	srcF := fmt.Sprintf("%v/%v/%v-source-%v", baseFileName, orch, mid, u[len(u)-1])
+	src, _, err := testers.SaveToExternalStorage(srcF, source)
+	if err != nil {
+		glog.Error(err)
+		return err
+	}
+	glog.Infof("Wrote segment to storage file=%v", src)
```
yondonfu, line 658: "segment to storage" -> "rendition to storage" to clarify which file is being saved?

yondonfu, line 658: file -> url to match the convention used in the next log statement below?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	}
+
+	if len(rendition) == 0 || len(source) == 0 {
+		return fmt.Errorf("no segments found")
+	}
+
+	y, m, d := start.Date()
+	baseFileName := fmt.Sprintf("%v-%v%v", y, m.String(), d)
+	u := strings.Split(renditionUrl, "/")
+	u = u[len(u)-2:]
```
yondonfu, line 650: Let's be more descriptive with var naming here so it is clearer what we are extracting from the slice of strings created in the above line.

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	for _, v := range sourcePl.Segments {
+		if v != nil {
+			sourceSegs = append(sourceSegs, v)
+		}
+	}
+	if len(rendSegs) == 0 || len(sourceSegs) == 0 {
+		return nil, nil, "", fmt.Errorf("no segments found")
+	}
+
+	seqNo := r % int(math.Min(float64(len(rendSegs)), float64(len(sourceSegs))))
```
yondonfu, line 706: Instead of constructing the extra `rendSegs` and `sourceSegs` slices, you should be able to just do:

```
segIdx := r % int(math.Min(float64(len(renditionPl.Segments)), float64(len(sourcePl.Segments)))
seqNo := sourcePl.Segments[segIdx].SeqId // Assumes that segIdx points to the same seqNo in the slice of sources and renditions
```

yondonfu, line 706: Since `r` isn't used until this line let's move its declaration closer to here?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+
+	pl, err := m3u8.NewMediaPlaylist(1, 1)
+	if err != nil {
+		return nil, err
+	}
+	if err := pl.Decode(*bytes.NewBuffer(body), true); err != nil {
+		return nil, err
+	}
+
+	for _, s := range pl.Segments {
```
yondonfu, line 801: When would there be a nil MediaSegment?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
+		return nil, errors.New(string(body))
+	}
+
+	pl, err := m3u8.NewMediaPlaylist(1, 1)
```
yondonfu, line 793: This will always return a windowed playlist with a single segment. Should we use the same [live list length](https://github.com/livepeer/go-livepeer/blob/14b0bd6d92f4b840ffb1e16b3d438640648e36a6/core/playlistmanager.go#L18) as the broadcaster?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+
+	if len(rendition) == 0 || len(source) == 0 {
+		return fmt.Errorf("no segments found")
+	}
+
+	y, m, d := start.Date()
+	baseFileName := fmt.Sprintf("%v-%v%v", y, m.String(), d)
+	u := strings.Split(renditionUrl, "/")
+	u = u[len(u)-2:]
+	rendF := fmt.Sprintf("%v/%v/%v-%v", baseFileName, orch, mid, strings.Join(u, "-"))
```
yondonfu, line 651: Format -> `date/orch/manifestID/rendition/seqNo.ts` to follow a similar file path convention as the B?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	srcF := fmt.Sprintf("%v/%v/%v-source-%v", baseFileName, orch, mid, u[len(u)-1])
```
yondonfu, line 652: Format -> `date/orch/manifestID/source/seqNo.ts` to follow a similar file path convention as the B?

yondonfu (changes requested):
Noting a few small differences from the desc in livepeer/livepeer-infra#315:

1. A random sample is taken for each O for *every* test instead of just for one test a day
2. The source + rendition for the sample could be different for each O instead of for any given test sampling the same source and renditions for each O

1 should be fine for now as we'll just end up with more data that we need - we can always just choose from the data available and having more is not a bad thing.

2 should be fine for now since the test videos are short and even if different sources + renditions are chosen someone visually inspecting should be able to fairly easily determine if content matches up. Might be an issue if test videos are longer, but since this is just a temporary thing I think its ok for now.

We may be able to avoid hardcoded a # of segments (which of course would vary if the test video length varied) if we end up switching to a full playlist (as mentioned in this [comment](https://github.com/livepeer/livepeer-infra/issues/326#issuecomment-747827945)) so just noting as something to potentially follow up on.

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +599,209 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
 
 	return data.Data.Transcoders, nil
 }
+
+// Get a random source/rendition sample segment and write it to external storage
+func (s *streamerClient) randomSample(ctx context.Context, mid string, orch string) error {
+	randSeqNo := rand.Int() % numSegments
+	randProfile := s.profiles[rand.Int()%len(s.profiles)]
```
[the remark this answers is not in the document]
kyriediculous: The profiles are defined first as a flag `presets` which is passed into the HTTP request for the stream-tester which I believe uses those presets for the transcoding. But they are not used on the broadcaster because the `-authWebhookUrl` is not defined on the broadcaster, right ? So yes in that case they would need to be the same.

cmd/orch-tester/orch_tester.go
kyriediculous: We currently retry until the stream finishes which would log the error on each retry , that might be a little bit verbose as it's not a critical error (we have a fallback)

cmd/orch-tester/orch_tester.go
kyriediculous: The loops filter out the nil values

cmd/orch-tester/orch_tester.go
kyriediculous: On the line below we decode the buffer into the playlist instance, wouldn't this expand the window size to what the buffer contains and since this would be the broadcaster's window size of 6 we will use that ?

cmd/orch-tester/orch_tester.go
kyriediculous: I felt like this would make the directory structure deeper than necessary , it would create two more folders for both the source and the rendition to only contain those files respectively.

kyriediculous:
Fixed most comments in c8c3676 and left a comment on others

cmd/orch-tester/orch_tester.go
```diff
@@ -184,6 +225,15 @@ func main() {
 			glog.Error(err)
 			continue
 		}
+		// if we haven't found a random sample by now cancel and wait for the backup attempt to complete before returning
+		if *randomSample {
+			cancel()
+			<-randErr
+			if err != nil {
```
yondonfu, line 232: This handles the err var defined outside of this conditional, but we want to handle the error returned from the `randErr` channel.

cmd/orch-tester/orch_tester.go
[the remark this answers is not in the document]
yondonfu: Got it. Yeah the only way to set the presets on the broadcaster is to either directly set them (i.e. via CLI, startup flags, etc.) or via a webhook. I believe the presets passed to stream-tester can be used for .com API stream configurations, but in this case since orch-tester is using the broadcaster directly I don't think the presets passed to stream-tester are actually used. So, the primary purpose of passing presets to orch-tester is for selecting the rendition to sample from an orchestrator.

cmd/orch-tester/orch_tester.go
kyriediculous: fixed in e0ee351

cmd/orch-tester/orch_tester.go
yondonfu: Ah and:

- `downloadSegment()` is supposed to return an error when the segment is not available yet
- `Verbose` type returned by `glog.V` only exposes Info based logging

I see - ok to do this for now then.

I do think that this is a bit of a code smell since an error is not included in the error log. If someone were to filter through ERROR logs, these errors would not show up. And if it's not a big deal for these errors not to show up that may be a prompt to consider whether that error is really an error (or if there is room to distinguish an expected scenario from a true error condition). Anyways, I understand why we're doing this right now, but thought this would be worth mentioning.

cmd/orch-tester/orch_tester.go
yondonfu: Bump.

cmd/orch-tester/orch_tester.go
yondonfu: Bump on this question. I see your other [comment](https://github.com/livepeer/stream-tester/pull/68#discussion_r546906111) about filtering out nil values, but I'd like to understand what would cause nil values in the middle of the slice.

kyriediculous: Right, I'm not sure if there would be any in the middle of the slice but I have encoutered a few nil values at the end of the slice. So if we're positive this doesn't occur in the middle of the slice either I guess we can break once we encounter a nil value.

cmd/orch-tester/orch_tester.go
kyriediculous: removed in f8c6438

cmd/orch-tester/orch_tester.go
yondonfu: I'm not sure. Could you confirm that this is the case by doing a simple test?

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +603,211 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	}
+
+	if len(rendition) == 0 || len(source) == 0 {
+		return fmt.Errorf("no segments found")
+	}
+
+	y, m, d := start.Date()
+	baseFileName := fmt.Sprintf("%v-%v-%v", y, m, d)
+	urlSplit := strings.Split(renditionUrl, "/")
+	urlSplit = urlSplit[len(urlSplit)-2:]
```
yondonfu, line 653: Would it be clearer to do something like:

```
rend := urlSplit[len(urlSplit)-2]
fname := urlSplit[len(urlSplit)-1]
```

And then combine the two together as needed in the following lines?

cmd/orch-tester/orch_tester.go
darkdarkdragon: Not, it will not. To decode manifest with unknown length this should be used:
https://github.com/livepeer/stream-tester/blob/bfc1bf03f93a003f58913154ba8a945cf2a1177e/internal/testers/m3utester2.go#L563-L574

cmd/orch-tester/orch_tester.go
yondonfu: I see. If it is possible for nil values to be in the middle of the slice then keeping the nil checks makes sense.

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +603,211 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	if err := pl.Decode(*bytes.NewBuffer(body), true); err != nil {
+		return nil, err
+	}
+
+	for _, s := range pl.Segments {
+		if s == nil {
+			break
```
darkdarkdragon, line 809: Still don't understand what is the point of that loop?
Also, there shouldn't be nil sigments in the middle of slice - you see nil segments because whole slice is allocated at creation, but not necessarily filled when decoded.
If you need number of segments in the playlist you should call `pl.Len()`

kyriediculous: Not sure either, I think it lingered from earlier. Will remove.

cmd/orch-tester/orch_tester.go
kyriediculous: This returns a `Playlist` type instead of a `MediaPlaylist` type though how can we get a `MediaPlaylist` ? 
Decode it line by line ?

darkdarkdragon: https://github.com/livepeer/stream-tester/blob/bfc1bf03f93a003f58913154ba8a945cf2a1177e/internal/testers/m3utester2.go#L959-L966

cmd/orch-tester/orch_tester.go
```diff
@@ -549,3 +603,219 @@ func (s *streamerClient) getOrchestrators() ([]*orch, error) {
...
+		return nil, nil, "", err
+	}
+
+	if len(renditionPl.Segments) == 0 {
+		return nil, nil, "", fmt.Errorf("no transcoded segments")
+	}
+
+	var rendSegs []*m3u8.MediaSegment
+	var sourceSegs []*m3u8.MediaSegment
+	for _, v := range renditionPl.Segments {
```
yondonfu, line 704: Since we can fetch the length of the playlist via the `Len()` method as mentioned in this [comment](https://github.com/livepeer/stream-tester/pull/68#discussion_r546942815) by @darkdarkdragon, I think we can omit these loops and make the following update on L721:

```
segIdx := r % int(math.Min(float64(renditionPl.Len()), float64(sourcePl.Len()))
```

which should ensure that we never get a nil segment (that might be at the end of the slice stored by the playlist) when indexing into the Segments slice.

kyriediculous:
I believe all changes are addresses, I also did a manual test and it's still good 👍

yondonfu (approved):
LGTM after squashing

PR #74, opened 2021-01-06
Orch-tester: fix random sampling

### LINKED ISSUE #73: orch-tester: Incorrect uploaded source + rendition when randomly sampling from playlists
When the orch-tester randomly samples source + rendition pairs to upload to GCP, I've observed source + rendition pairs where:

- The rendition contains what appears to be a source video
- The source contains what appears to be a rendition video, but it is not the rendition indicated by the filename i.e. if the filename indicates a 360p video, the rendition is not a 360p video

[Example incorrect source](https://console.cloud.google.com/storage/browser/_details/leaderboard-samples/2021-January-4/0x525419ff5707190389bfb5c87c375d710f5fcb0e/bbbm3u8_3d9e0d43c0568c7400f2-source-11.ts?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=test-harness-226018)
[Example incorrect rendition](https://console.cloud.google.com/storage/browser/_details/leaderboard-samples/2021-January-4/0x525419ff5707190389bfb5c87c375d710f5fcb0e/bbbm3u8_3d9e0d43c0568c7400f2-P360p30fps16x9-11.ts?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=test-harness-226018)

Upon looking at the orch-tester random sampling code, I think there are 2 bugs that are triggered when we sample from the playlists (as opposed to choosing a random segment and trying to download the source + rendition while the stream is in-progress).

- sampleFromPlaylist() returns the [source as the first return value and the rendition as the second value](https://github.com/livepeer/stream-tester/blob/0254b402c62f2cf87580a5815b6f34e63fa72e37/cmd/orch-tester/orch_tester.go#L731), but when calling the function, the [first return value is stored as the rendition and the second return value is stored as the source](https://github.com/livepeer/stream-tester/blob/0254b402c62f2cf87580a5815b6f34e63fa72e37/cmd/orch-tester/orch_tester.go#L625).
- Turns out [this assumption](https://github.com/livepeer/stream-tester/blob/0254b402c62f2cf87580a5815b6f34e63fa72e37/cmd/orch-tester/orch_tester.go#L720) is actually invalid because the source playlist can contain all segments while the rendition playlist might only contain a subset (for the segments that were successfully transcoded. Suppose the index is 3. The 3rd segment in the source playlist would be segment 3, but the 3rd segment in the rendition playlist might be segment 5 (as an example to illustrate that this could be a number besides 3).

kyriediculous:
Fixes #73

### DIFF
--- a/cmd/orch-tester/orch_tester.go
+++ b/cmd/orch-tester/orch_tester.go
@@ -717,8 +717,7 @@ func (s *streamerClient) sampleFromPlaylist(mid string) (source, rendition []byt
 
 	r := rand.Int()
 	segIdx := r % int(math.Min(float64(renditionPl.Len()), float64(sourcePl.Len())))
-	seqNo := sourcePl.Segments[segIdx].SeqId
-	rendURI := renditionPl.Segments[seqNo].URI
+	rendURI := renditionPl.Segments[segIdx].URI
 	urlSplit := strings.Split(rendURI, "/")
 	fName := urlSplit[len(urlSplit)-1]
 
@@ -728,12 +727,12 @@ func (s *streamerClient) sampleFromPlaylist(mid string) (source, rendition []byt
 	}
 
 	sourceURI := fmt.Sprintf("%v/stream/%v_0_0/source/%v", s.broadcaster, mid, fName)
-	source, err = s.downloadSegment(fmt.Sprintf("%v/%v", s.broadcaster, sourceURI))
+	source, err = s.downloadSegment(sourceURI)
 	if err != nil {
 		return nil, nil, "", err
 	}
 
-	return source, rendition, renditionPl.Segments[seqNo].URI, err
+	return source, rendition, rendURI, err
 }
 
 func (s *streamerClient) downloadSegment(url string) ([]byte, error) {
--- a/cmd/orch-tester/orch_tester.go
+++ b/cmd/orch-tester/orch_tester.go
@@ -622,7 +622,7 @@ func (s *streamerClient) randomSample(ctx context.Context, mid string, orch stri
 		case <-ctx.Done():
 			glog.Info("random sampler timed out")
 			// If we haven't found the segment try to sample from whatever playlist we can
-			rendition, source, renditionUrl, err = s.sampleFromPlaylist(mid)
+			source, rendition, renditionUrl, err = s.sampleFromPlaylist(mid)
 			if err != nil {
 				glog.Error(err)
 			}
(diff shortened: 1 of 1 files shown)


cmd/orch-tester/orch_tester.go
```diff
@@ -717,13 +717,18 @@ func (s *streamerClient) sampleFromPlaylist(mid string) (source, rendition []byt
...
+	urlSplit := strings.Split(rendURI, "/")
+	fName := urlSplit[len(urlSplit)-1]
+
+	rendition, err = s.downloadSegment(fmt.Sprintf("%v/%v", s.broadcaster, rendURI))
 	if err != nil {
 		return nil, nil, "", err
 	}
 
-	source, err = s.downloadSegment(fmt.Sprintf("%v/%v", s.broadcaster, sourcePl.Segments[seqNo].URI))
+	sourceURI := fmt.Sprintf("%v/stream/%v_0_0/source/%v", s.broadcaster, mid, fName)
```
yondonfu, line 730: This URL structure doesn't seem right. Looks like `s.broadcaster` would be included twice in the end URL passed to `downloadSegment()`.

yondonfu, line 730: I see that we previously used `sourcePl.Segments[...].URI` and we're still using `renditionPl.Segments[...].URI`. Why can't we use `sourcePl.Segments[...].URI` here?

yondonfu (changes requested):

cmd/orch-tester/orch_tester.go
kyriediculous: We use the filename ('seqNo'.ts) of the rendition we selected to download its matching source since `sourcePl.Segments[index]` doesn't necessarily match `reditionPl.Segments[index]`

cmd/orch-tester/orch_tester.go
```diff
@@ -717,13 +717,18 @@ func (s *streamerClient) sampleFromPlaylist(mid string) (source, rendition []byt
 
 	r := rand.Int()
 	segIdx := r % int(math.Min(float64(renditionPl.Len()), float64(sourcePl.Len())))
-	seqNo := sourcePl.Segments[segIdx].SeqId // Assumes that segIdx points to the same seqNo in the slice of sources and renditions
-	rendition, err = s.downloadSegment(fmt.Sprintf("%v/%v", s.broadcaster, renditionPl.Segments[seqNo].URI))
+	seqNo := sourcePl.Segments[segIdx].SeqId
+	rendURI := renditionPl.Segments[seqNo].URI
```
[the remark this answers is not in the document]
kyriediculous: I believe this should be fixed by https://github.com/livepeer/stream-tester/pull/74#discussion_r553446996 and e135a68 cleans it up a little bit

yondonfu (approved):
LGTM after squashing. Let's verify this on staging and then if everything looks good update the prod deployment.

PR #75, opened 2021-02-09
mist-api-connector: add -consul-prefix, regex host

iameli:
Changes necessary for global playback URL — need the location prefixes on there so we can tell the difference between different datacenters, and need the regex to respond to both `{region}-playback-{secret}.livepeer.monster` and `playback-{secret}.livepeer.monster`.


darkdarkdragon (approved):
LGTM

PR #79, opened 2021-02-19
mist-api-connector: implement SRT support via PUSH_REWRITE

iameli:
Still needs a change to the `CONN_CLOSE` trigger, but this works for the new generic `PUSH_REWRITE` trigger that works for both RMTP and SRT.


iameli:
Superseded by https://github.com/livepeer/stream-tester/pull/97

PR #87, opened 2021-06-30
implement RTMP push

iameli:
The only tricky bit here - I currently have a hacky 10-second timeout to make sure transcoded renditions exist before starting stream pushes. Yuck. Should handle that better.

Should also maybe use their "autopush" functionality instead of the regular push so it auto-retries if we lose connection to the remote server.

### DIFF
--- a/apis/mist/mist.go
+++ b/apis/mist/mist.go
@@ -170,6 +170,17 @@ type (
 		StopSessions []string `json:"stop_sessions,omitempty"`
 		NukeStream   string   `json:"nuke_stream,omitempty"` // Stops running stream (forces RTMP to disconnect)
 	}
+
+	pushStartAction struct {
+		Params map[string]string `json:"params,omitempty"`
+		Stream string            `json:"stream,omitempty"`
+		Target string            `json:"target,omitempty"`
+	}
+
+	pushStartReq struct {
+		Authorize *authReq         `json:"authorize,omitempty"`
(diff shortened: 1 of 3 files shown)


apis/mist/mist.go
```diff
@@ -333,6 +344,26 @@ func (mapi *API) NukeStream(name string) error {
...
+func (mapi *API) StartPush(stream string, target string) error {
+	glog.Infof("Starting push for '%s'", stream)
+	reqs := &pushStartReq{
+		Authorize: &authReq{
+			Username: mapi.login,
+			Password: mapi.challengeRepsonse,
+		},
+		PushStart: &pushStartAction{
+			Params: map[string]string{
+				"video": "854x480",
```
darkdarkdragon, line 356: Looks like values here doesn't matter? Can we put comment about it if so?

darkdarkdragon (approved):
Small question, otherwise, looks good

About 10s pause before starting push - right way to do it will be to start pulling main mafinest in a loop (to check that correct rendition appeared), but we can do this in the next iteration.

apis/mist/mist.go
iameli: Ah yep - I thought maybe that was where the video selector goes. But it goes in the target query parameter.

victorges:
Just realized I had totally missed this PR 🥲
Gotta work out my github notification setup

PR #90, opened 2021-07-16
mist-api-connector: Recover etcd session when lease is lost

victorges:
This is a proposal for the `etcd` write failure we are still getting once and a while in production.

I believe it will be necessary even after we change the `etcd` endpoints to contain all members of 
the cluster, since `mist-api-connector` itself is already using the `AutoSyncInterval` option of
the client, which means that it will only stay attached to a single server on the first 5 minutes 
after startup. After that, the client will grab all endpoints from the cluster automatically.

So we are still losing the session sometimes for some reason, and I believe having a recovering
mechanism like this will fix any issues from that. Also added a last resort mechanic of shutting 
down the server in case we can't recover the session after 2 minutes, which feels like a time we
would be restarting the mist servers ourselves anyway.

### DIFF
--- a/internal/app/mistapiconnector/mistapiconnector_app.go
+++ b/internal/app/mistapiconnector/mistapiconnector_app.go
@@ -687,7 +687,7 @@ func newEtcdSession(etcdClient *clientv3.Client) (*concurrency.Session, error) {
 	if err != nil {
 		return nil, fmt.Errorf("mist-api-connector: Error creating etcd session err=%w", err)
 	}
-	glog.Info("etcd got lease %d", sess.Lease())
+	glog.Infof("etcd got lease %d", sess.Lease())
 	return sess, nil
 }
 
--- a/internal/app/mistapiconnector/mistapiconnector_app.go
+++ b/internal/app/mistapiconnector/mistapiconnector_app.go
@@ -690,11 +690,7 @@ func (mc *mac) recoverEtcdSessionOnce() error {
 }
 
 func newEtcdSession(etcdClient *clientv3.Client) (*concurrency.Session, error) {
-	ctx, cancel := context.WithTimeout(etcdClient.Ctx(), etcdDialTimeout)
-	defer cancel()
-	sess, err := concurrency.NewSession(etcdClient,
-		concurrency.WithTTL(etcdSessionTTL),
-		concurrency.WithContext(ctx))
+	sess, err := concurrency.NewSession(etcdClient, concurrency.WithTTL(etcdSessionTTL))
 	if err != nil {
 		return nil, fmt.Errorf("mist-api-connector: Error creating etcd session err=%w", err)
 	}
--- a/internal/app/mistapiconnector/mistapiconnector_app.go
+++ b/internal/app/mistapiconnector/mistapiconnector_app.go
@@ -636,7 +636,7 @@ func (mc *mac) recoverSessionLoop() {
 		cancel()
 
 		if err != nil && clientCtx.Err() == nil {
-			glog.Error("Shutting down due to unrecoverable etcd session. err=%q. ", err)
+			glog.Errorf("Shutting down due to unrecoverable etcd session. err=%q.", err)
 			mc.shutdown()
 			return
 		}
--- a/internal/app/mistapiconnector/mistapiconnector_app.go
+++ b/internal/app/mistapiconnector/mistapiconnector_app.go
@@ -89,7 +89,7 @@ type (
 		useEtcd        bool
 		etcdClient     *clientv3.Client
 		etcdSession    *concurrency.Session
-		etcdPub2rev    map[string]etcdRevData // public key to revision of ETCD keys
+		etcdPub2rev    map[string]etcdRevData // public key to revision of etcd keys
 	}
 )
 
@@ -126,7 +126,7 @@ func NewMac(mistHost string, mapi *mist.API, lapi *livepeer.API, balancerHost st
 			DialOptions:      []grpc.DialOption{grpc.WithBlock()},
 		})
 		if err != nil {
-			err = fmt.Errorf("mist-api-connector: Error connecting ETCD err=%w", err)
+			err = fmt.Errorf("mist-api-connector: Error connecting etcd err=%w", err)
 			return nil, err
 		}
 		sess, err = newEtcdSession(cli)
@@ -150,7 +150,7 @@ func NewMac(mistHost string, mapi *mist.API, lapi *livepeer.API, balancerHost st
 		useEtcd:        useEtcd,
 		etcdClient:     cli,
 		etcdSession:    sess,
-		etcdPub2rev:    make(map[string]etcdRevData), // public key to revision of ETCD keys
+		etcdPub2rev:    make(map[string]etcdRevData), // public key to revision of etcd keys
 		srvShutCh:      make(chan error),
 	}
 	go mc.recoverSessionLoop()
@@ -505,7 +505,7 @@ func (mc *mac) handleDefaultStreamTrigger(w http.ResponseWriter, r *http.Request
 		}
 	}
 	if mc.useEtcd {
-		// now create routing rule in the ETCD for HLS playback
+		// now create routing rule in the etcd for HLS playback
 		if mc.baseStreamName != "" {
 			wildcardPlaybackID := mc.wildcardPlaybackID(stream)
 			playbackID := mc.consulPrefix + stream.PlaybackID
@@ -546,7 +546,7 @@ func (mc *mac) handleDefaultStreamTrigger(w http.ResponseWriter, r *http.Request
 			)
 		}
 		if err != nil {
-			glog.Errorf("Error creating ETCD Traefik rule for playbackID=%s streamID=%s err=%v", stream.PlaybackID, stream.ID, err)
+			glog.Errorf("Error creating etcd Traefik rule for playbackID=%s streamID=%s err=%v", stream.PlaybackID, stream.ID, err)
 			w.WriteHeader(http.StatusInternalServerError)
 			w.Write([]byte("false"))
 		}
@@ -578,7 +578,7 @@ func (mc *mac) putEtcdKeys(sess *concurrency.Session, playbackID string, kvs ...
 	if !resp.Succeeded {
 		panic("unexpected")
 	}
-	glog.Infof("for playbackID=%s created %d keys in ETCD revision=%d", playbackID, len(kvs)/2, resp.Header.Revision)
+	glog.Infof("for playbackID=%s created %d keys in etcd revision=%d", playbackID, len(kvs)/2, resp.Header.Revision)
 	mc.etcdPub2rev[playbackID] = etcdRevData{resp.Header.Revision, kvs}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -122,14 +129,12 @@ func NewMac(mistHost string, mapi *mist.API, lapi *livepeer.API, balancerHost st
 			err = fmt.Errorf("mist-api-connector: Error connecting ETCD err=%w", err)
 			return nil, err
 		}
-		sess, err = concurrency.NewSession(cli, concurrency.WithTTL(etcdSessionTTL))
+		sess, err = newEtcdSession(cli)
```
victorges, line 132: Just realized we can also force a `cli.Sync()` immediately here, so we don't have to wait for the auto-sync interval before the client picks all cluster endpoints. That only if we still might take some time until having the full member list in the endpoints configuration.

That said, also came to my mind that the session might be lost sometimes because we currently only _create_ by the time we have a single endpoints in the `etcd` client? Than the 3 endpoints could be the thing that fixes the session issues, but it still feels like something that could happen anytime since the session is bound to a manual renewal of a time-limited lease. Also not sure if the `etcd` protocol requires that you send the writes to all members of the cluster for it to really have an effect, sounds like something the `etcd` servers would handle in between themselves.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -551,15 +558,15 @@ func (mc *mac) handleDefaultStreamTrigger(w http.ResponseWriter, r *http.Request
...
+func (mc *mac) putEtcdKeys(sess *concurrency.Session, playbackID string, kvs ...string) error {
 	if len(kvs) == 0 || len(kvs)%2 != 0 {
 		return errors.New("number of arguments should be even")
 	}
 	cmp := clientv3.Compare(clientv3.CreateRevision(kvs[0]), ">", -1) // basically noop - will always be true
 	thn := make([]clientv3.Op, 0, len(kvs)/2)
 	get := clientv3.OpGet(kvs[0])
 	for i := 0; i < len(kvs); i += 2 {
-		thn = append(thn, clientv3.OpPut(kvs[i], kvs[i+1], clientv3.WithLease(mc.etcdSession.Lease())))
+		thn = append(thn, clientv3.OpPut(kvs[i], kvs[i+1], clientv3.WithLease(sess.Lease())))
```
victorges, line 569: I am wondering if, at least when "recovering" the keys, we should have a different `clientv3.Compare` function here, like to only set the `etcd` keys iff the current key mod (create?) revision is `0` (doesn't exist?) or exactly what the revision we had before (to put it simply: `modRev < myRev+1`).

Is it possible that when one `mist-api-connector` instance loses the session/lease, the stream would immediately go to some `mist-api-connector` instance which would create different routes for it?

darkdarkdragon: > Is it possible that when one mist-api-connector instance loses the session/lease, the stream would immediately go to some mist-api-connector instance which would create different routes for it?

When mist-api-connector looses session it doesn't mean that connection stopped. When connection stopped Mist sends CLOSE_CONN trigger to mist-api-connector. So only case when connection is stopped but mist-api-connector is not aware about that - is when Mist crashes, which hasn't happened for us for a long time.

internal/app/mistapiconnector/mistapiconnector_app.go
darkdarkdragon: > Just realized we can also force a cli.Sync() immediately here,

Let's do it.

>  Also not sure if the etcd protocol requires that you send the writes to all members

No, client sends to one server, and then servers handle that between themselves.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -610,8 +617,78 @@ func (mc *mac) deleteEtcdKeys(playbackID string) {
...
+
+func (mc *mac) recoverSessionLoop() {
+	clientCtx := mc.etcdClient.Ctx()
+	for clientCtx.Err() == nil {
+		select {
+		case <-clientCtx.Done():
+			// client closed, which means app shutted down
+			return
+		case <-mc.etcdSession.Done():
+		}
```
darkdarkdragon, line 632:
```suggestion
		}
		glog.Infof("etcd session with lease=%d is lost, trying to recover", mc.etcdSession.Lease())
```

darkdarkdragon (changes requested):

darkdarkdragon:
@victorges 
> Also added a last resort mechanic of shutting
down the server in case we can't recover the session after 2 minutes, 

Do we know what k8s will do in such case? mist-api-connector runs inside one pod with Mist. In case mist-api-connector exits - will k8s restart whole pod or only mist-api-connector's container? Because if Mist is not restarted then there is no point in restarting mist-api-connector - streams will continue running but there will no be etcd routes for them...

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -610,8 +617,78 @@ func (mc *mac) deleteEtcdKeys(playbackID string) {
...
 	}
+
+	mc.etcdSession.Close()
+	mc.etcdSession = sess
+	glog.Infof("Recovered etcd session. lease=%d", sess.Lease())
+	return nil
+}
+
+func newEtcdSession(etcdClient *clientv3.Client) (*concurrency.Session, error) {
+	sess, err := concurrency.NewSession(etcdClient, concurrency.WithTTL(etcdSessionTTL))
```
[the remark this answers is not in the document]
victorges: Actually found a way to enforce a timeout to this call with the `concurrency.WithContext` option. Did that instead. 
Do you think we still need the log?

victorges:
@darkdarkdragon 
>Do we know what k8s will do in such case? mist-api-connector runs inside one pod with Mist. In case mist-api-connector exits - will k8s restart whole pod or only mist-api-connector's container? Because if Mist is not restarted then there is no point in restarting mist-api-connector - streams will continue running but there will no be etcd routes for them...

Ugh, I always forget that k8s restarts specific containers and not the whole pod when they crash. You are right, and only now I understood why that "crash the process if session lost" idea wasn't such a quick win after all...

I did some research and thought about our options here, and I think they are:
 - Have some logic in mac to actually call k8s API to delete the pod when the session is unrecoverable (my least favorite);
 - Share process namespace between the containers and submit a direct SIGINT/KILL to Mist here (not very fond of that either);
 - Have some direct interaction via the Mist API to force Mist server to also "shutdown" in some way, like at least requesting to stop all the current streams; 
 - Have a liveness probe in Mist Server that somehow fails if the mistapiconnector is not healthy/connected? No idea how that could be done though (is the only/best way to learn about mist through those [manuals](https://mistserver.org/guides/MistServer_Manual_2.18.pdf)?)
 - (Also) have a readiness probe in `mac`, so that at least the mist server won't receive **new** sessions until the etcd session is ready (and we would see it in kubectl)

All of those seem like out of scope from this PR anyway, so in here I'll just remove this forced shutdown. 
Instead, I will add a log when the session is taking too long to be recovered, and:
 - In the first moment, we can keep that "etcd write failure" alarm to monitor it and see if it self-heals fine;
 - later, when we trust this new logic, we change that alarm to only catch this log for when the session is unrecoverable, so we can explicitly restart the pods ourselves.
WDYT?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -571,8 +585,8 @@ func (mc *mac) putEtcdKeys(playbackID string, kvs ...string) error {
 	if !resp.Succeeded {
 		panic("unexpected")
 	}
-	glog.Infof("for playbackID=%s created %d keys in ETCD revision=%d", playbackID, len(kvs)/2, resp.Header.Revision)
-	mc.etcdPub2rev[playbackID] = resp.Header.Revision
+	glog.Infof("for playbackID=%s created %d keys in etcd revision=%d", playbackID, len(kvs)/2, resp.Header.Revision)
+	mc.etcdPub2rev[playbackID] = etcdRevData{resp.Header.Revision, kvs}
```
victorges, line 589: Another thing that came to my mind: What happens if we get a write failure above and never get to save the `etcd` entries in memory here? Does the actual stream on Mist fail or does it start without the etcd entries? If the latter, should we "backup" the etcd entries in `etcdPub2rev` first thing in this function and here only add the revision number? (properly handling an invalid revision number like `-1` wherever we use it, like in the delete below)

internal/app/mistapiconnector/mistapiconnector_app.go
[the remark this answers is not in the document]
darkdarkdragon: Nope, will not do it.  Same context gets used in the call to `client.KeepAlive(ctx, id)`, so the moment this context is canceled KeepAlive loop is canceled too and session is lost.
Only way is to copy whole `session.go` to our repo and modify it.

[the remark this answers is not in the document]
victorges: Oh you're right it uses the same ctx for creating the initial lease and for the keep alive loop, sorry I missed that.
I'll just revert and make the log you suggested then.

internal/app/mistapiconnector/mistapiconnector_app.go
darkdarkdragon: If `putEtcdKeys` call fails then stream will be denied.
BTW, I've spotted small bug - please add fix for it 😄 
```go
		if err != nil {
			glog.Errorf("Error creating ETCD Traefik rule for playbackID=%s streamID=%s err=%v", stream.PlaybackID, stream.ID, err)
			w.WriteHeader(http.StatusInternalServerError)
			w.Write([]byte("false")) // here we tell Mist to deny stream.
// fixme: forgot to return here
		}
	}
	w.Write([]byte(responseURL))
	metrics.StartStream()
	glog.Infof("Responded with '%s'", responseURL)
```

darkdarkdragon:
@victorges 
> All of those seem like out of scope from this PR anyway, so in here I'll just remove this forced shutdown.
Instead, I will add a log when the session is taking too long to be recovered, and:

>    In the first moment, we can keep that "etcd write failure" alarm to monitor it and see if it self-heals fine;
    later, when we trust this new logic, we change that alarm to only catch this log for when the session is unrecoverable, so we can explicitly restart the pods ourselves.
    If that unrecoverable session actually starts happening too frequently as well (which I'm hoping not), we can come back to this discussion and maybe consider any of the workarounds above (or some other bigger/better change).

Given that `NewSession` call probably will be blocked until connection is restored (though I don't know what errors can production etcd return - I've tested only locally with one instance of etcd) - probably we will never hit `unrecoverable` failure.
I think for now we can just monitor `etcd session with lease=%d is lost, trying to recover"` and `"mist-api-connector: unrecoverable etcd session. err=%q."` messages - I hope that code you've already wrote will solve all our problems with etcd and we will not be returning to those problems.

internal/app/mistapiconnector/mistapiconnector_app.go
victorges: [Fixed!](https://github.com/livepeer/stream-tester/pull/90/commits/fd32d4ff96d82fca2622e43f745969fa36a490ec)

darkdarkdragon (approved):
LGTM

PR #91, opened 2021-07-27
mist-api-connector: Implement healthcheck for k8s probes

victorges:
This is first to implement a healthcheck API in mapic server, to allow creating a readiness probe in k8s.

On the way to that also found out that we have frequent restarts (up to 2) on `mist-api-connector` because
it tries to connect too fast to Mist, before it is ready. So I implemented a little retry logic on startup so we 
give some time for Mist server to be available before giving up and restarting the container.

Some idea I had for the future is being able to configure mapic containers with a `restartPolicy: never`, so
we can make sure that the mapic container never restarts. Those would create inconsistencies between its
state in etcd/memory with what is actually happening on mist side, so we probably want some datadog alerts
in case any of those happen (either mist or mapic)

### DIFF
--- a/internal/app/mistapiconnector/mistapiconnector_app.go
+++ b/internal/app/mistapiconnector/mistapiconnector_app.go
@@ -1039,12 +1039,30 @@ func (mc *mac) createMistStream(streamName string, stream *livepeer.CreateStream
 	return err
 }
 
+func (mc *mac) handleHealthcheck(w http.ResponseWriter, r *http.Request) {
+	if !mc.isEtcdSessionHealthy() {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (mc *mac) isEtcdSessionHealthy() bool {
+	select {
+	case <-mc.etcdSession.Done():
+		return false
+	default:
+		return true
+	}
+}
+
 func (mc *mac) webServerHandlers() *http.ServeMux {
 	mux := http.NewServeMux()
 	utils.AddPProfHandlers(mux)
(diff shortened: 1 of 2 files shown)


internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -739,12 +739,30 @@ func (mc *mac) createMistStream(streamName string, stream *livepeer.CreateStream
...
+	if !mc.isEtcdSessionHealthy() {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (mc *mac) isEtcdSessionHealthy() bool {
+	select {
+	case <-mc.etcdSession.Done():
```
darkdarkdragon, line 752: this will crash if mapic used without ETCD

darkdarkdragon (changes requested):

internal/app/mistapiconnector/mistapiconnector_app.go
victorges: Fixed.

Turns out that the background `recoverSessionLoop` was also broken, and would panic mapic just after starting. I didn't know that running it without an etcd was actually still a thing. Do we have any of those happening in produciton right now? (cause they're likely to be broken due to the recover loop)

darkdarkdragon: > Do we have any of those happening in produciton right now? (cause they're likely to be broken due to the recover loop)

No, but ETCD is an still optional in mapic, so we should keep it working without it - or make ETCD mandatory.

darkdarkdragon (approved):
LGTM after rebase

PR #92, opened 2021-07-29
mist-api-connector: Check mist-api on healthcheck

victorges:
Do so by calling GetConfig API which felt lightweight enough
to be called frequently (every ~5s or so)


darkdarkdragon (approved):
LGTM

PR #93, opened 2021-07-29
mist-api-connector: Use AMQP Producer from livepeer-data/pkg/event

victorges:
Notice that a couple of changes were necessary in the usage of the producer
due to some changes in the implementation after merging the existing versions
of the AMQP producer.

This commit contains the exact changes compared to the version that existed in
this repository: https://github.com/livepeer/livepeer-data/pull/1/commits/8bd5ed33fcf42c612c38bbaf8f5c1048c67af34c


darkdarkdragon (approved):
LGTM

PR #94, opened 2021-08-06
mist-api-connector: Rename (rtmp)push(target) to multistream(target)

victorges:
Following the [rename on the API](https://github.com/livepeer/livepeer-com/pull/557), this does the same renames
on the current only known client for the feature.


iameli (approved):
LGTM

darkdarkdragon (approved):
LGTM

PR #96, opened 2021-08-18
mist-api-connector: Use webhook event types from livepeer-data

victorges:
Has some updated formats, following https://github.com/livepeer/livepeer-com/pull/569


internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -736,37 +736,29 @@ func (mc *mac) waitPush(info *streamInfo, pushInfo *pushStatus) {
...
-		UserID:    info.stream.UserID,
-		StreamID:  info.stream.ID,
-		Payload: map[string]interface{}{
-			"target": map[string]interface{}{
-				"id":      pushInfo.target.ID,
-				"name":    pushInfo.target.Name,
-				"profile": pushInfo.profile,
-			},
+
+	payload := data.MultistreamWebhookPayload{
```
victorges, line 744: It would be really nice to add some information about what is the error here, in case of a `multistream.error` event. Otherwise, all we'll be able to say is that the Multistream target hasn't connected, but no info about why.

@darkdarkdragon do you think we can grab that information from mist somehow?

darkdarkdragon (approved):
LGTM

internal/app/mistapiconnector/mistapiconnector_app.go
victorges: Talked on planning about this. From a product perspective, let's stick with what we have for this initial launch and have some directions about what to do in the docs instead. We can work on surfacing these error messages on the next iteration.

PR #103, opened 2021-09-01
mist-api-connector: add router priority

### LINKED ISSUE #99: traefik routing: newer streams should have a higher priority
We should set Traefik's "Priority" field to the current Unix timestamp. That way newer streams will override over streams, and double routes (even though they shouldn't happen) won't be a problem

iameli:
fixes #99

Today I learned about Go time formatting!


darkdarkdragon (approved):
LGTM

PR #106, opened 2021-09-16
add /cmaf routing, remove consul support

iameli:
Added new routes for /cmaf in addition to /hls. Moved all the key-generation stuff out into a helper function with conditionals. Did this actually make things clearer? Who knows!

Also removed Consul support 'cause we had a bad experience and I don't think we're going to support it no more.


cmd/mist-api-connector/mist-api-connector.go
```diff
@@ -52,6 +50,9 @@ func main() {
 	amqpUrl := fs.String("amqp-url", "", "RabbitMQ url")
 	_ = fs.String("config", "", "config file (optional)")
 
+	consulPrefix := fs.String("consul-prefix", "", "DEPRECATED - use --route-prefix")
+	consulMistURL := fs.String("consul-mist-url", "", "DEPRECATED - use --route-mist-url")
```
victorges, line 54: Nice! Should we also have a `--consul` flag like `DEPRECATED - consul is not supported anymore`?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -443,6 +429,54 @@ func (mc *mac) triggerDefaultStream(w http.ResponseWriter, r *http.Request, line
...
+		traefikKeyPathServices + serviceName + "/loadbalancer/servers/0/url",
+		mc.mistURL,
+		traefikKeyPathServices + serviceName + "/loadbalancer/passhostheader",
+		"false",
+	}
+
+	// Add a millisecond timestamp as the priority so new streams always win over old, stale streams
+	priority := strconv.FormatInt(time.Now().UnixNano()/int64(time.Millisecond), 10)
+
+	for _, prefix := range []string{"hls", "cmaf"} {
```
victorges, line 447: **nit:** Maybe have a "const" for this slice like `var playbackProtoPrefixes = []string{"hls", "cmaf"}` or something like that? (later could even become cli flags as well? 🤔)

cmd/mist-api-connector/mist-api-connector.go
```diff
@@ -40,10 +39,9 @@ func main() {
 	sendAudio := fs.String("send-audio", "record", "when should we send audio?  {always|never|record}")
 	apiToken := fs.String("api-token", "", "Token of the Livepeer API to be used by the Mist server")
 	apiServer := fs.String("api-server", livepeer.ACServer, "Livepeer API server to use")
-	consulURI := fs.String("consul", "", "Base URL to access Consul (for example: http://localhost:8500)")
-	consulPrefix := fs.String("consul-prefix", "", "Prefix to be prepended to all created consul routes e.g. 'nyc-'")
-	playbackDomain := fs.String("playback-domain", "", "regex of domain to create consul routes for (ex: playback.livepeer.live)")
-	mistURL := fs.String("consul-mist-url", "", "external URL of this Mist instance (to be put in Consul) (ex: https://mist-server-0.livepeer.live)")
+	routePrefix := fs.String("route-prefix", "", "Prefix to be prepended to all created routes e.g. 'nyc-'")
+	playbackDomain := fs.String("playback-domain", "", "regex of domain to create routes for (ex: playback.livepeer.live)")
```
victorges, line 43: Since this is a regex, should it be like `playback\.livepeer\.live`? Guess it doesn't matter much in the example, but I wonder if we set it correctly in prod

victorges (commented):
Can't say much about the etcd rules/traefik behavior itself, but assuming the idea was just to duplicate everything `/hls/` into `/cmaf/` as well, lgtm!

cmd/mist-api-connector/mist-api-connector.go
iameli: Trafeik's regex thingy is really weird and not well documentedd, I think only things that are within curly braces get parsed as regexes? ["HostRegexp and Path accept an expression with zero or more groups enclosed by curly braces. Named groups can be like {name:pattern} that matches the given regexp pattern or like {name} that matches anything until the next dot."](https://doc.traefik.io/traefik/routing/routers/) So our current prod value is `{prefix:.*}{domain:(playback-redacted|cdn)}.livepeer.com` and I believe that works as intended

cmd/mist-api-connector/mist-api-connector.go
iameli: If it gets run with that it'll crash with `flag provided but not defined: -consul` — good enough for me, pretty sure that has zero users right now

internal/app/mistapiconnector/mistapiconnector_app.go
iameli: Done - originally tried to do it with a `const` and it didn't work, didn't occur to me to just make a package-scoped `var` 🤦‍♂️

PR #107, opened 2021-09-24
CI: Publish version tags with all images

victorges:
Following up on https://github.com/livepeer/livepeer-infra/pull/686, we would like to have some version tags
in all of our repos. This is the first repo after `livepeer-data` being updated to this new format.

Differently from `livepeer-data`, I haven't really repeated [the whole logic](https://github.com/livepeer/livepeer-data/compare/v0.2.1...v0.2.4#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R3) of grabbing the containing
branch for a tag build, nor have I enabled `tag` event builds for our docker images (only the release workflow
is actually run on those).

I will try to create a github action that extracts all the raw tags that we want to publish for a given image to
avoid repating that script everywhere else.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -17,20 +17,30 @@ jobs:
       env:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
           DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-      run: docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD 
+      run: docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+    - name: Get git tag
+      id: git_tag
+      run: |
+        echo ::set-output name=VERSION::$(git describe --tag --dirty)
+    - name: Get build tag
+      id: build_tag
+      run: echo ::set-output name=VERSION::${GITHUB_REF#refs/tags/}
+
     - name: Build the Docker image
-      # run: docker build . --file Dockerfile --tag livepeer/streamtester:$(date +%s)
-      run: docker build . --file Dockerfile --tag livepeer/streamtester:latest --build-arg version=$(git describe --dirty)
-    - name: Get tag name
-      id: get_version
-      run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+      run: docker build . --file Dockerfile --tag livepeer/streamtester:latest --build-arg version=${{ steps.git_tag.outputs.VERSION }}
+
+    - name: Push Docker Container to Registry
+      run: docker push livepeer/streamtester:latest
     - name: Tag version
       if: startsWith(github.ref, 'refs/tags/v')
       run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/load-tester-dockerimage.yml
```diff
@@ -17,17 +17,22 @@ jobs:
...
-    - name: Get tag name
-      id: get_version
-      run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
-    - name: Tag version
-      if: startsWith(github.ref, 'refs/tags/v')
       run: |
-        docker tag livepeer/streamtester:load-tester livepeer/streamtester:load-tester${{ steps.get_version.outputs.VERSION }}
-        docker push livepeer/streamtester:load-tester${{ steps.get_version.outputs.VERSION }}
+        TAGS=(${{ steps.tags.outputs.tags }})
+        docker build . --file Dockerfile.load-tester ${TAGS[@]/#/--tag livepeer/streamtester:} --build-arg "version=${{ steps.tags.outputs.version }}"
```
iameli, line 32: This... is not easy to Google. It's like a bash inline foreach...? Looks cool.

.github/workflows/dockerimage.yml
```diff
@@ -17,20 +17,24 @@ jobs:
       env:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
           DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-      run: docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD 
+      run: docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+
+    - name: Tags
+      id: tags
+      uses: livepeer/action-gh-release-tags@v0
```
iameli, line 24: `v0` here is a tag we can just keep updating to move these forward, yeah? It's not pinned or something?

iameli (changes requested):
I've created `livepeer/mist-api-connector`, `livepeer/loadtester`, and `livepeer/record-tester`. Wanna set those up here?

.github/workflows/load-tester-dockerimage.yml
victorges: It is! Got it from here: https://stackoverflow.com/questions/20366609/prefix-and-postfix-elements-of-a-bash-array/30971572

Maybe better to switch to the `sh`-compatible way of doing this, as I did it on `livepeer-com`? Something like `$(printf ' -t livepeer/streamtester:%s' $TAGS)`

.github/workflows/dockerimage.yml
victorges: Yeah! So we can just move the tag on the action repository and it will start using the latest version it's pointing to ✨

.github/workflows/load-tester-dockerimage.yml
victorges: Done!

iameli (approved):

PR #108, opened 2021-09-26
mist-api-connector: fix deleteEtcdKeys

iameli:
Broken because it didn't have the right prefix. But a small change gets it working again.

Verified working in `livepeer-in-a-box`.


iameli:
Merging this without review to stave off an impending prod meltdown 😬

PR #110, opened 2021-10-01
mist-api-connector: Add retries to `/setactive` request

victorges:
This is to add a couple retries to the `/setactive` call to the API, to hopefully diminish the inconsistencies
we observe on some streams that are actually active but the state on the database say they are not.

Also added a log that we should create an alert on.

### DIFF
--- a/apis/livepeer/livepeer.go
+++ b/apis/livepeer/livepeer.go
@@ -3,6 +3,7 @@ package livepeer
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


victorges:
FTR: Measured `/setactive` API latency from the logs and it seems that the latency P99 stays at around 1.2s (on the last 24h), so thats why a timeout of 1.5s.

https://docs.google.com/spreadsheets/d/12VNYlysQLUBUPY0CyakayInltM5CZ4nUGqpgSbWwFJ8/edit?usp=sharing

iameli (approved):
LGTM. (But let's still figure out a way to make the analyzer answer the "isActive question someday.)

victorges:
> @iameli let's still figure out a way to make the analyzer answer the "isActive question someday

Yeah definitely! We will get that event flowing ASAP for multi-region already, then it's just a matter of creating a simple reducer in the analyzer to hold that information. I actually want to create it pretty soon as well, to fix some inconsistencies that show up in the dashboard after a stream has been stopped and restarted. e.g. we currently have no way of knowing whether a "disconnected" state or something like that is from the current session or a previous one.

PR #112, opened 2021-10-20
mist-api-connector: Send `stream.state` event with active/inactive stream info

victorges:
This is built on top of https://github.com/livepeer/stream-tester/pull/104 to also start sending an event with some simpler "stream state" to be
processed by the `analyzer`. 

This event is also the main one **to be replicated globally** and create the distributed stream state on
the `analyzer` instances, leveraging the RabbitMQ Federation created on https://github.com/livepeer/livepeer-infra/pull/707

The event is currently only sent when the stream state "knowingly" changed like when it became active or
inactive, but we should already think about implementing some kind of "keep-alive" logic, so that we don't 
get zombie-active streams that are kept as active forever because the Mist instance has gone away.

This is also built on top of features in https://github.com/livepeer/livepeer-data/pull/18

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/gosuri/uiprogress v0.0.1
 	github.com/livepeer/joy4 v0.1.2-0.20210601043311-c1b885884cc7
 	github.com/livepeer/leaderboard-serverless v1.0.0
-	github.com/livepeer/livepeer-data v0.4.0-beta
+	github.com/livepeer/livepeer-data v0.4.0
 	github.com/livepeer/m3u8 v0.11.1
 	github.com/mattn/go-isatty v0.0.8 // indirect
 	github.com/patrickmn/go-cache v2.1.0+incompatible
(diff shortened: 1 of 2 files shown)


internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -691,40 +693,52 @@ func (mc *mac) waitPush(info *streamInfo, pushInfo *pushStatus) {
...
 
-func (mc *mac) emitWebhookEvent(info *streamInfo, pushInfo *pushStatus, eventKey string) {
-	if mc.producer == nil {
-		return
+func (mc *mac) emitStreamStateEvent(stream *livepeer.CreateStreamResp, state data.StreamState) {
+	streamID, sessionID := stream.ParentID, stream.ID
+	if streamID == "" {
+		streamID = sessionID
 	}
+	stateEvt := data.NewStreamStateEvent(streamID, stream.UserID, sessionID, state)
```
victorges, line 706: Note to self: need to add the region to this event as well.

PR #117, opened 2021-11-17
.github/ci: Build images on every branch

victorges:
Use the `always-latest-on-branch` input to the gh action (created [here](https://github.com/livepeer/action-gh-release-tags/commit/1442c82d03ecfc7adf61e5c4c8c796bda167812b))
so we only publish `latest` tags on the specified branches.

While `latest` will only be pushed on those branches, we can still test the
generated images from any branch since I removed the action filter to only
run on master/mist-api-connector branch.

Also thinking if I should get rid of the `mist-api-connector` branch. Will
create a discussion inline.

### DIFF
--- a/.github/workflows/mist-api-connector-dockerimage.yml
+++ b/.github/workflows/mist-api-connector-dockerimage.yml
@@ -16,8 +16,6 @@ jobs:
     - name: Tags
       id: tags
       uses: livepeer/action-gh-release-tags@v0
-      with:
-        always-latest-on-branch: mist-api-connector
 
     - name: Build the Docker image
       run: |
--- a/.github/workflows/mist-api-connector-dockerimage.yml
+++ b/.github/workflows/mist-api-connector-dockerimage.yml
@@ -1,14 +1,8 @@
 name: Mist API Connector Docker Image CI
-
-on:
-  push:
-    branches:
-      - mist-api-connector
+on: push
 
 jobs:
-
   build:
-
     runs-on: ubuntu-latest
 
     steps:
@@ -23,7 +17,7 @@ jobs:
       id: tags
       uses: livepeer/action-gh-release-tags@v0
       with:
-        force-latest: true
+        always-latest-on-branch: mist-api-connector
 
     - name: Build the Docker image
       run: |
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -1,14 +1,8 @@
 name: Docker Image CI
-
-on:
-  push:
-    branches:
-      - master
+on: push
 
 jobs:
-
   build:
-
     runs-on: ubuntu-latest
 
     steps:
@@ -23,7 +17,7 @@ jobs:
       id: tags
       uses: livepeer/action-gh-release-tags@v0
       with:
-        force-latest: true
+        always-latest-on-branch: master
 
     - name: Build the Docker image
       run: |
--- a/.github/workflows/load-tester-dockerimage.yml
+++ b/.github/workflows/load-tester-dockerimage.yml
@@ -1,14 +1,8 @@
 name: Load Tester Docker Image CI
-
-on:
-  push:
-    branches:
-      - master
+on: push
 
 jobs:
-
   build:
-
     runs-on: ubuntu-latest
 
     steps:
@@ -23,7 +17,7 @@ jobs:
       id: tags
       uses: livepeer/action-gh-release-tags@v0
       with:
-        force-latest: true
+        always-latest-on-branch: master
 
     - name: Build the Docker image
       run: |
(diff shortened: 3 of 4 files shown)


.github/workflows/mist-api-connector-dockerimage.yml
```diff
@@ -23,7 +17,7 @@ jobs:
       id: tags
       uses: livepeer/action-gh-release-tags@v0
       with:
-        force-latest: true
+        always-latest-on-branch: mist-api-connector
```
victorges, line 20: @darkdarkdragon WDYT of removing the `mist-api-connector` branch now that we?:
 - Already have separate images for every version of this repository, like `livepeer/record-tester`, `livepeer/stream-tester`, `livepeer/mist-api-connector`, etc (so no tag conflicts between them)
 - Will be building and publishing images from every branch after this change, so one could easily test a dev version by using the tag with the branch name (or git tag name), e.g. [livepeer/mist-api-connector:vg-chore-simplify-ci](https://hub.docker.com/layers/livepeer/mist-api-connector/vg-chore-simplify-ci/images/sha256-64449afcd29b49f6cf68da5f7f80b10cc975bc695fd0ecde1983258b431bd629?context=explore)

victorges: In the current configuration, we will only tag images with `latest` in case the build comes from the `mist-api-connector` branch. If we change this to `master`, any code pushed to `master` will be automatically built and deployed to staging. They won't go to prod until we make an explicit deploy though since we don't auto-deploy to prod as of now, but could cause accidental deploys in case one doesn't check the diff before `actually_deploy`ing.

To have that extra layer of security for deploys from staging to prod, one thing that we could do is: 
 - Remove this `always-latest-on-branch:` completely from this action. This will use [the default behavior of the action](https://github.com/livepeer/action-gh-release-tags/commit/1442c82d03ecfc7adf61e5c4c8c796bda167812b#diff-d3b71acd6af9d3cffc47fe34c80c23636c9f5e4d2610d8ca7fdc9f4d9ba961c5L29) which only tags with `latest` for **stable** semantic versions (so pre-releases like `v1.2.3-beta` are _not_ latest)
 - Change values in `livepeer-infra`, so that prod keeps using `latest`, but make `staging` start using the "trunk" version of the image i.e. `mist-api-connector:master`

This will force us to always need to tag a version with a stable semver before it can start being used by production code. I actually like that behavior, if you agree with it I'll probably update some other services to follow it too 😁 
cc @iameli too, WDYT?

darkdarkdragon: @darkdarkdragon 
> This will force us to always tag a version with a stable semver before the image is updated in

This feels unclear.
I would want clear conscious manual action that shows intention to release. Tagging something is not being associated with such intention (at least in my mind).

I want this layers of protection because `magic` is crucial in our infrastructure and accidental release of wrong version will bring down all the production clusters at once.

victorges: > I would want clear conscious manual action that shows intention to release. Tagging something is not being associated with such intention (at least in my mind).

I am used to working on projects where tagging was exactly this intent-ful action for making a release (sometimes an automatic deploy as well!). Given that context, in my mind it is more associated with a release than pushing to a branch (which I generally associate more with "staging/next" versions since it's more unstable/mutable).

Another alternative would be if we did not use the `latest` tag in `livepeer-infra` for `prod` (which would automatically resolve to the last stable semver pushed), but instead always specify a complete tagged version [here](https://github.com/livepeer/livepeer-infra/blob/6839f96a9d1b99cc4f32c81a47f893a981a20c04/values/mist.values.yaml#L50).

This would add an additional step to the deploy process though, which is to also update that reference with the new semver, instead of only rebasing the branch or only tagging a stable semver. Whole thing:
 - Tag version in the repository here
 - Go to `livepeer-infra`, explicitly update the version reference (instead of having the automatic image update)

WDYT?

darkdarkdragon: @victorges 

> Another alternative would be if we did not use the latest tag in livepeer-infra for prod (which would automatically resolve to the last stable semver pushed), but instead always specify a complete tagged version here.


This is exactly what I've had in mind.

victorges:
Cool! Will do that then, I think it's an appropriate flow as well for such
critical service. Multi-factor deploy 😄

.github/workflows/mist-api-connector-dockerimage.yml
victorges: Do you think we should follow that multi-step pattern for every image built in the repository, or only for mapic which is more critical?

darkdarkdragon: `magic` is the only image that is used in production and as such even has 'deploy' process. So, I think, it makes sense to just always build everything else.

PR #118, opened 2021-11-19
record-tester: Test stream health

victorges:
This creates the `stream_health.go` tester, to validate that the stream health becomes
available during the `record-tester` tests. 

Uses the analyzer client created on https://github.com/livepeer/livepeer-data/pull/27

This implements: https://github.com/livepeer/livepeer-data/issues/25


cmd/recordtester/recordtester.go
```diff
@@ -65,6 +68,7 @@ func main() {
 		ff.WithConfigFileFlag("config"),
 		ff.WithConfigFileParser(ff.PlainParser),
 		ff.WithEnvVarPrefix("RT"),
+		ff.WithEnvVarIgnoreCommas(true),
```
iameli, line 71: 😤

PR #121, opened 2021-12-20
mist-api-connector: Export metrics for multistream usage minutes/bytes

victorges:
They will be globally aggregated on Victoria Metrics and retained for at least 30 days, so
that we can create with a view for the total number of multistreamed minutes (and bytes)
per month.

This implements https://github.com/livepeer/livepeer-com/issues/675


iameli (approved):
LGTM

PR #129, opened 2022-01-28
record-tester: Allow recording streams to custom object store

victorges:
- Allow setting custom `recordObjectStoreId` on API client (pending: porting this change to https://github.com/livepeer/go-api-client/pull/1)
 - Receive the OS ID as a CLI flag and pass it around until the place we actually create the streams

### DIFF
--- a/apis/livepeer/livepeer.go
+++ b/apis/livepeer/livepeer.go
@@ -64,9 +64,10 @@ type (
 		} `json:"servers,omitempty"`
 	}
 
-	createStreamReq struct {
-		Name    string   `json:"name,omitempty"`
-		Presets []string `json:"presets,omitempty"`
+	CreateStreamReq struct {
+		Name     string   `json:"name,omitempty"`
+		ParentID string   `json:"parentId,omitempty"`
+		Presets  []string `json:"presets,omitempty"`
 		// one of
 		// - P720p60fps16x9
 		// - P720p30fps16x9
@@ -77,8 +78,9 @@ type (
 		// - P240p30fps16x9
 		// - P240p30fps4x3
 		// - P144p30fps16x9
-		Profiles []Profile `json:"profiles,omitempty"`
-		Record   bool      `json:"record,omitempty"`
+		Profiles            []Profile `json:"profiles,omitempty"`
+		Record              bool      `json:"record,omitempty"`
+		RecordObjectStoreId string    `json:"recordObjectStoreId,omitempty"`
 	}
 
 	// Profile transcoding profile
@@ -346,15 +348,6 @@ var StandardProfiles = []Profile{
 	},
 }
 
-// CreateStream creates stream with specified name and profiles
-func (lapi *API) CreateStream(name string, presets ...string) (string, error) {
-	csr, err := lapi.CreateStreamEx(name, false, presets)
-	if err != nil {
-		return "", err
-	}
-	return csr.ID, err
-}
-
 // DeleteStream deletes stream
 func (lapi *API) DeleteStream(id string) error {
 	glog.V(model.DEBUG).Infof("Deleting Livepeer stream '%s' ", id)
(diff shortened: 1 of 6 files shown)


victorges:
Will keep this one open since it might be useful for testing Storj.

apis/livepeer/livepeer.go
```diff
@@ -383,36 +376,27 @@ func (lapi *API) DeleteStream(id string) error {
 
 // CreateStreamEx creates stream with specified name and profiles
 func (lapi *API) CreateStreamEx(name string, record bool, presets []string, profiles ...Profile) (*CreateStreamResp, error) {
-	return lapi.CreateStreamEx2(name, record, "", presets, profiles...)
+	return lapi.CreateStream(CreateStreamReq{Name: name, Record: record, Presets: presets, Profiles: profiles})
 }
 
-// CreateStreamEx creates stream with specified name and profiles
-func (lapi *API) CreateStreamEx2(name string, record bool, parentID string, presets []string, profiles ...Profile) (*CreateStreamResp, error) {
```
tqian1, line 390: Looks like we are getting rid of these, great they were pretty confusing - do we need to check if this will break builds elsewhere?

apis/livepeer/livepeer.go
```diff
@@ -64,9 +64,10 @@ type (
 		} `json:"servers,omitempty"`
 	}
 
-	createStreamReq struct {
-		Name    string   `json:"name,omitempty"`
-		Presets []string `json:"presets,omitempty"`
+	CreateStreamReq struct {
+		Name     string   `json:"name,omitempty"`
+		ParentID string   `json:"parentId,omitempty"`
```
tqian1, line 69: How come ParentID wasn't here before?

apis/livepeer/livepeer.go
```diff
@@ -383,36 +376,27 @@ func (lapi *API) DeleteStream(id string) error {
 
 // CreateStreamEx creates stream with specified name and profiles
 func (lapi *API) CreateStreamEx(name string, record bool, presets []string, profiles ...Profile) (*CreateStreamResp, error) {
```
tqian1, line 378: Curious how come this func is named CreateStreamEx, what does `Ex` mean or what is the naming convention here?

figintern (commented):
Excited for this, had some questions about it from hackers

Just had a few questions

apis/livepeer/livepeer.go
victorges: I wouldn't be much worried about it since the build will only reaaally break when they do the `go get -u` on this package. What I'm a little more worried is that a little later, for VOD, I created this separate repository to hold the API client instead of having everyone depending on this internal package: https://github.com/livepeer/go-api-client

So these changes should be upstreamed there and we should eventually, hopefully soon, migrate this repository to use that lib instead of having it's own implementation of a Livepeer API client. That other repository started exactly from this code tho, so it should not be very hard to migrate anyway.

apis/livepeer/livepeer.go
victorges: So the `parentID` was actually only added to the path of the creation request like this: 
https://github.com/livepeer/stream-tester/pull/129/files#diff-df99bcc770f1faed4b227511eaadf4caeab0940378f2046e2e2d526176a6acf9L414

It is not a problem to add the field in the request body though, since when we are setting it with the same value as the value set in the path, and also because in the end it will be just ignored: the child stream will have its `parentId` field [coming from the parent stream](https://github.com/livepeer/livepeer-com/blob/dffc577067cfc88c2c32876f2ff28491c77494c8/packages/api/src/controllers/stream.ts#L721), which is just fetched from the [`streamId` (parent ID) param in the path](https://github.com/livepeer/livepeer-com/blob/dffc577067cfc88c2c32876f2ff28491c77494c8/packages/api/src/controllers/stream.ts#L625).

So I am OK with keeping this field in the `POST` request, also because it is also abiding to the `stream` object schema anyway, so even if we do that validation sometime it will continue to be valid.

If you think it would be better no to include this field in the request payload though, it is really easy to do so as well. I can just change the `json` tag here to be `json:"-"` instead. WDYT?

apis/livepeer/livepeer.go
victorges: I don't really know TBH, it is before my time. I could only guess it that means something like "extended"? Like we first had `CreateStream` but then wanted some extension of it with additional capabilities. I know that the `R` suffix means "with retries", so you might find some of those with that.

apis/livepeer/livepeer.go
tqian1: Right makes sense, will keep an eye out for using `go-api-client`

apis/livepeer/livepeer.go
tqian1: I don't see a prob with including the field in the payload, just wanted to confirm what it is for.

figintern (commented):
LGTM

figintern (approved):

PR #131, opened 2022-02-04
feature: mist-api-connector tags per-stream metrics

tqian1:
* This change adds streamID tags to the `multistream_usage_megabytes` `multistream_usage_minutes` metrics of mist-api-connector service

Testing:
compiles with `make connector`

Issue:
https://app.zenhub.com/workspaces/video-user-team-5dd2c75e4ee08500015994d7/issues/livepeer/go-livepeer/2172

### DIFF
--- a/internal/app/mistapiconnector/stats_collector.go
+++ b/internal/app/mistapiconnector/stats_collector.go
@@ -95,11 +95,11 @@ func createMetricsEvent(nodeID, region string, info *streamInfo, metrics *stream
 				MediaTimeMs: push.Stats.MediaTime,
 			}
 			if metrics.Bytes > pushInfo.pushedBytes {
-				census.IncMultistreamBytes(metrics.Bytes-pushInfo.pushedBytes, info.id)
+				census.IncMultistreamBytes(metrics.Bytes-pushInfo.pushedBytes, info.stream.PlaybackID) // manifestID === playbackID
 				pushInfo.pushedBytes = metrics.Bytes
 			}
 			if mediaTime := time.Duration(metrics.MediaTimeMs) * time.Millisecond; mediaTime > pushInfo.pushedMediaTime {
-				census.IncMultistreamTime(mediaTime-pushInfo.pushedMediaTime, info.id)
+				census.IncMultistreamTime(mediaTime-pushInfo.pushedMediaTime, info.stream.PlaybackID)
 				pushInfo.pushedMediaTime = mediaTime
 			}
 		}
--- a/internal/app/mistapiconnector/stats_collector.go
+++ b/internal/app/mistapiconnector/stats_collector.go
@@ -25,9 +25,6 @@ type metricsCollector struct {
 }
 
 func startMetricsCollector(ctx context.Context, period time.Duration, nodeID, ownRegion string, mapi *mist.API, producer *event.AMQPProducer, amqpExchange string, infop infoProvider) {
-	census.IncMultistreamBytes(0, "")
-	census.IncMultistreamTime(0, "")
-
 	mc := &metricsCollector{nodeID, ownRegion, mapi, producer, amqpExchange, infop}
 	go mc.mainLoop(ctx, period)
 }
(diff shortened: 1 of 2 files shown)


internal/app/mistapiconnector/stats_collector.go
```diff
@@ -25,8 +25,8 @@ type metricsCollector struct {
 }
 
 func startMetricsCollector(ctx context.Context, period time.Duration, nodeID, ownRegion string, mapi *mist.API, producer *event.AMQPProducer, amqpExchange string, infop infoProvider) {
-	census.IncMultistreamBytes(0)
```
tqian1, line 28: Not sure why/if this initialization of 0 bytes is needed. Since there is no streamID at this point we are pushing the zero value string `""`

darkdarkdragon: I think in this case initialisation should be removed because it creates weird datapoint with empty streamID.

internal/app/mistapiconnector/stats_collector.go
```diff
@@ -98,11 +98,11 @@ func createMetricsEvent(nodeID, region string, info *streamInfo, metrics *stream
 				MediaTimeMs: push.Stats.MediaTime,
 			}
 			if metrics.Bytes > pushInfo.pushedBytes {
-				census.IncMultistreamBytes(metrics.Bytes - pushInfo.pushedBytes)
+				census.IncMultistreamBytes(metrics.Bytes-pushInfo.pushedBytes, info.stream.ID)
```
darkdarkdragon, line 101: I think we should use `info.stream.PlaybackID` (which we call manifestID in other places).
@victorges @iameli wdyt?

tqian1: is `PlaybackID`/`manifestID` the ID that appears in the livepeer.com dashboard? ie. `ad73b342-2c64-4a51-af55-47e830ba1d05`

internal/app/mistapiconnector/stats_collector.go
tqian1: That sounds good, any reason it required the initialization before?

internal/app/mistapiconnector/stats_collector.go
darkdarkdragon: stream.ID it is in dashboard:
image
stream.PlaybackID - it is part of Playback URL:
image
(https://cdn.livepeer.monster/hls/ef6aah6kn7zzdaxd/index.m3u8)
And stream.PlaybackID in our logs gets logged as `manifestID`

darkdarkdragon: And most log lines contains `manifestID`, no `stream.ID`

internal/app/mistapiconnector/stats_collector.go
darkdarkdragon: Because without initialisation Grafana doesn't show anything for the metric, so it's hard to tell if metric just not been measured yet or there is some problem in scraping.

internal/app/mistapiconnector/stats_collector.go
```diff
@@ -98,11 +95,11 @@ func createMetricsEvent(nodeID, region string, info *streamInfo, metrics *stream
 				MediaTimeMs: push.Stats.MediaTime,
 			}
 			if metrics.Bytes > pushInfo.pushedBytes {
-				census.IncMultistreamBytes(metrics.Bytes - pushInfo.pushedBytes)
+				census.IncMultistreamBytes(metrics.Bytes-pushInfo.pushedBytes, info.id)
```
darkdarkdragon, line 98: Could we use `info.stream.PlaybackID` here?

internal/metrics/census.go
```diff
@@ -70,6 +71,7 @@ func InitCensus(nodeID, version, namespace string) {
 	var err error
 	ctx := context.Background()
 	Census.kNodeID = tag.MustNewKey("node_id")
+	Census.kStreamID = tag.MustNewKey("stream_id")
```
darkdarkdragon, line 74: Could we change this to be `manifest_id` for consistency with `go-livepeer`?

darkdarkdragon (changes requested):

PR #134, opened 2022-02-17
vod: Allow video manipulation logic to be used as a library

darkdarkdragon:
changes needed for VOD

Will allow using stream-tester internal abstractions both on the transcode CLI as well as the VOD task-runner.


victorges (approved):
LGTM

PR #136, opened 2022-03-21
recordtester: Add ingest arg to skip LP /api/ingest

victorges:
That API now returns the global ingest and we don't want that.
We should instead stream into the region being tested and so
the best way for that should be receiving an explicit argument
instead, since we don't want to expose the regional URLs in any
way to our users.

### DIFF
--- a/messenger/messenger.go
+++ b/messenger/messenger.go
@@ -6,8 +6,10 @@ package messenger
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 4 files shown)


messenger/messenger.go
```diff
@@ -316,30 +320,14 @@ func sendMessage(msg string) {
...
-				msg1 := msg[:l]
-				msg = msg[l:]
-				sendMessage(msg1)
-				if len(msg) == 0 {
-					break
-				}
-			}
-			return
-		}
-	*/
```
iameli, line 342: Why commented?

iameli (approved):

messenger/messenger.go
victorges: It was commented before and exactly the same code that is present in the other `sendMessage` function (this is `sendMessageRaw`). I assumed it was just moved there at some point but the commented code stayed here, so I removed it.

PR #137, opened 2022-03-31
workflow: Rewrite workflow actions + codesign and notarize builds

hjpotter92:
- Setup release pipeline in similar vein to that of <https://github.com/livepeer/go-livepeer>
- Add a test workflow job
- Setup macos releases to be notarized through Apple servers before cutting release
- Add video-user group as codeowner for project

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - master
-      - "hjp/*"
 
 jobs:
   linux:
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,214 @@
...
+        id: go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
+
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          # Check https://github.com/livepeer/go-livepeer/pull/1891
```
iameli, line 29: love this

PR #140, opened 2022-04-08
mapic: Added triggerUserNew

gioelecerati:
**What does this pull request do? Explain your changes. (required)**

Catches USER_NEW trigger from Mist and call the user configured webhooks on the user.new event (probably to be renamed).
Currently it responds `false` to Mist if one or more webhooks fail.
Depends on [#995](https://github.com/livepeer/livepeer-com/pull/995) and [#994](https://github.com/livepeer/livepeer-com/pull/994).

**Specific updates (required)**

- Added triggerUserNew
- Added UserWebhook struct to `lapi`
- Added GetWebhooksByUserId on `lapi`

**How did you test each of these updates (required)**

livepeer-in-a-nice-box

### DIFF
--- a/apis/livepeer/livepeer.go
+++ b/apis/livepeer/livepeer.go
@@ -462,11 +462,11 @@ func (lapi *API) DefaultPresets() []string {
 }
 
 // GetWebhooksForEvent gets webhooks for an event by user id
-func (lapi *API) GetWebhooksForEvent(userId string, event string) ([]UserWebhook, error) {
+func (lapi *API) GetWebhooksForEvent(userId string, streamId string, event string) ([]UserWebhook, error) {
 	if userId == "" || event == "" {
 		return nil, fmt.Errorf("userId and event must be specified")
 	}
-	u := fmt.Sprintf("%s/api/webhook/subscribed/%s?userId=%s", lapi.choosenServer, event, userId)
+	u := fmt.Sprintf("%s/api/webhook/subscribed/%s?userId=%s&streamId=%s", lapi.choosenServer, event, userId, streamId)
 	return lapi.GetWebhooks(u)
 }
 
--- a/apis/livepeer/livepeer.go
+++ b/apis/livepeer/livepeer.go
@@ -134,14 +134,21 @@ type (
 
 	// User Webhooks
 	UserWebhook struct {
-		ID           string   `json:"id,omitempty"`
-		Url          string   `json:"url,omitempty"`
-		Kind         string   `json:"kind,omitempty"`
-		Name         string   `json:"name,omitempty"`
-		Events       []string `json:"events,omitempty"`
-		UserID       string   `json:"userId,omitempty"`
-		CreatedAt    int64    `json:"createdAt,omitempty"`
-		SharedSecret string   `json:"sharedSecret,omitempty"`
+		ID           string        `json:"id,omitempty"`
+		Url          string        `json:"url,omitempty"`
+		Kind         string        `json:"kind,omitempty"`
+		Name         string        `json:"name,omitempty"`
+		Events       []string      `json:"events,omitempty"`
+		UserID       string        `json:"userId,omitempty"`
+		CreatedAt    int64         `json:"createdAt,omitempty"`
+		SharedSecret string        `json:"sharedSecret,omitempty"`
+		Status       WebhookStatus `json:"status,omitempty"`
+	}
+
+	// Webhooks status
+	WebhookStatus struct {
+		LastTriggeredAt int64                  `json:"lastTriggeredAt,omitempty"`
+		LastFailure     map[string]interface{} `json:"lastFailure,omitempty"`
 	}
 
 	MultistreamTarget struct {
@@ -468,6 +475,12 @@ func (lapi *API) GetWebhooksForEvent(userId string, streamId string, event strin
 	}
 	u := fmt.Sprintf("%s/api/webhook/subscribed/%s?userId=%s&streamId=%s", lapi.choosenServer, event, userId, streamId)
 	return lapi.GetWebhooks(u)
+}
+
+// Update webhook status & response
+func (lapi *API) UpdateWebhookStatus(id string, status map[string]interface{}) bool {
+	go lapi.UpdateWebhook(id, status)
+	return true
 }
 
 // GetStreamByKey gets stream by streamKey
@@ -818,6 +831,50 @@ func (lapi *API) GetWebhooks(u string) ([]UserWebhook, error) {
 	}
 
 	return r, nil
+}
+
+func (lapi *API) UpdateWebhook(id string, status map[string]interface{}) error {
+	rType := "update_webhook"
+	start := time.Now()
+	u := fmt.Sprintf("%s/api/webhook/%s/status", lapi.choosenServer, id)
+
+	b, err := json.Marshal(status)
+	if err != nil {
+		glog.Errorf("Error marshalling webhook status to json: %v", err)
+		metrics.APIRequest(rType, 0, err)
+		return err
+	}
+	req, err := uhttp.NewRequest("POST", u, bytes.NewBuffer(b))
+
+	if err != nil {
+		glog.Errorf("Error creating webhook status request: %v", err)
+		metrics.APIRequest(rType, 0, err)
+		return err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+lapi.accessToken)
+	req.Header.Add("Content-Type", "application/json")
+	resp, err := lapi.httpClient.Do(req)
+	if err != nil {
+		glog.Errorf("Error updating webhook from Livepeer API server (%s) error: %v", u, err)
+		metrics.APIRequest(rType, 0, err)
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusNoContent {
+		b, _ := ioutil.ReadAll(resp.Body)
+		glog.Errorf("Status error updating webhook from Livepeer API server (%s) status %d body: %s", u, resp.StatusCode, string(b))
+		if resp.StatusCode == http.StatusNotFound {
+			metrics.APIRequest(rType, 0, ErrNotExists)
+			return ErrNotExists
+		}
+		err := errors.New(http.StatusText(resp.StatusCode))
+		metrics.APIRequest(rType, 0, err)
+		return err
+	}
+	took := time.Since(start)
+	metrics.APIRequest(rType, took, nil)
+	return nil
 }
 
 func (lapi *API) GetMultistreamTarget(id string) (*MultistreamTarget, error) {
--- a/apis/livepeer/livepeer.go
+++ b/apis/livepeer/livepeer.go
@@ -134,21 +134,34 @@ type (
 
 	// User Webhooks
 	UserWebhook struct {
-		ID           string        `json:"id,omitempty"`
-		Url          string        `json:"url,omitempty"`
-		Kind         string        `json:"kind,omitempty"`
-		Name         string        `json:"name,omitempty"`
-		Events       []string      `json:"events,omitempty"`
-		UserID       string        `json:"userId,omitempty"`
-		CreatedAt    int64         `json:"createdAt,omitempty"`
-		SharedSecret string        `json:"sharedSecret,omitempty"`
-		Status       WebhookStatus `json:"status,omitempty"`
+		ID           string   `json:"id,omitempty"`
+		Url          string   `json:"url,omitempty"`
+		Kind         string   `json:"kind,omitempty"`
+		Name         string   `json:"name,omitempty"`
+		Events       []string `json:"events,omitempty"`
+		UserID       string   `json:"userId,omitempty"`
+		CreatedAt    int64    `json:"createdAt,omitempty"`
+		SharedSecret string   `json:"sharedSecret,omitempty"`
 	}
 
-	// Webhooks status
-	WebhookStatus struct {
-		LastTriggeredAt int64                  `json:"lastTriggeredAt,omitempty"`
-		LastFailure     map[string]interface{} `json:"lastFailure,omitempty"`
+	WebhookUpdateStatusRequest struct {
+		ErrorMessage string                `json:"errorMessage,omitempty"`
+		Response     WebhookResponseStatus `json:"response,omitempty"`
+	}
+
+	WebhookResponseStatus struct {
+		ID         string          `json:"webhookId,omitempty"`
+		CreatedAt  int64           `json:"createdAt,omitempty"`
+		StatusCode int             `json:"statusCode,omitempty"`
+		Response   WebhookResponse `json:"response,omitempty"`
+	}
+
+	WebhookResponse struct {
+		Body       string            `json:"body"`
+		Status     int               `json:"status,omitempty"`
+		Headers    map[string]string `json:"headers,omitempty"`
+		Redirected bool              `json:"redirected,omitempty"`
+		StatusText string            `json:"statusText,omitempty"`
 	}
 
 	MultistreamTarget struct {
@@ -478,9 +491,8 @@ func (lapi *API) GetWebhooksForEvent(userId string, streamId string, event strin
 }
 
 // Update webhook status & response
-func (lapi *API) UpdateWebhookStatus(id string, status map[string]interface{}) bool {
+func (lapi *API) UpdateWebhookAsync(id string, status WebhookUpdateStatusRequest) {
 	go lapi.UpdateWebhook(id, status)
-	return true
 }
 
 // GetStreamByKey gets stream by streamKey
@@ -833,7 +845,7 @@ func (lapi *API) GetWebhooks(u string) ([]UserWebhook, error) {
 	return r, nil
 }
 
-func (lapi *API) UpdateWebhook(id string, status map[string]interface{}) error {
+func (lapi *API) UpdateWebhook(id string, status WebhookUpdateStatusRequest) error {
 	rType := "update_webhook"
 	start := time.Now()
 	u := fmt.Sprintf("%s/api/webhook/%s/status", lapi.choosenServer, id)
--- a/apis/livepeer/livepeer.go
+++ b/apis/livepeer/livepeer.go
@@ -132,6 +132,18 @@ type (
 		Mp4Url          string `json:"mp4Url,omitempty"`
 	}
 
+	// User Webhooks
+	UserWebhook struct {
+		ID           string   `json:"id,omitempty"`
+		Url          string   `json:"url,omitempty"`
+		Kind         string   `json:"kind,omitempty"`
+		Name         string   `json:"name,omitempty"`
+		Events       []string `json:"events,omitempty"`
+		UserID       string   `json:"userId,omitempty"`
+		CreatedAt    int64    `json:"createdAt,omitempty"`
+		SharedSecret string   `json:"sharedSecret,omitempty"`
+	}
+
 	MultistreamTarget struct {
 		ID        string `json:"id,omitempty"`
 		URL       string `json:"url,omitempty"`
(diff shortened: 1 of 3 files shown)


internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,6 +473,77 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+	}
+
+	userId := stream.UserID
+	userWebhooks, err := mc.lapi.GetWebhooksByUserId(userId, "user.new")
+	if err != nil {
+		glog.Errorf("Error getting webhooks for user %s err=%v", userId, err)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("false"))
+		return false
+	}
```
victorges, line 538: Same as above, I think we can even avoid calling this function every time a new user starts playing as well. Instead, we could call `GetWebhooksByUserId` as soon as the stream starts and save it into the `streamInfo` struct as well. 

i.e. Would add a `userNewWebhooks` field in `streamInfo` struct and initialize it with the response from this API when we create a stream info [here](https://github.com/livepeer/stream-tester/blob/fce88570dc0c3e7f86c86e4a8076e996b58d0880/internal/app/mistapiconnector/mistapiconnector_app.go#L546).

That way we avoid **all** I/O while handling this hook apart from the inevitable ones (calling the webhooks themselves). Since we are having lots of network issues in our clusters could be a good idea. WDYT?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,6 +473,77 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			glog.Errorf("Error reading response body from user.new webhook %s err=%v", userWebhook.Url, err)
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("false"))
+			return false
+		}
+
+		if resp.StatusCode/100 != 2 {
```
victorges: Never seen a status code check like this 😆

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,6 +473,77 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+	}
+
+	if len(userWebhooks) == 0 {
+		glog.V(model.DEBUG).Infof("No user.new webhooks for user %s", userId)
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("false"))
+		return false
+	}
+
+	for _, userWebhook := range userWebhooks {
```
victorges, line 547: We should probably call these in parallel (or not, if we only expect users to have 1 webhook anyway). It's easy to do it in Go though with an `errgroup`, [check it out](https://pkg.go.dev/golang.org/x/sync/errgroup).

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,6 +473,77 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+			w.Write([]byte("false"))
+			return false
+		}
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			glog.Errorf("Error reading response body from user.new webhook %s err=%v", userWebhook.Url, err)
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("false"))
+			return false
```
victorges: I think we should err on the side of permissiveness here. If the webhook is experiencing errors, we should allow the playback to go through by default. This will avoid affecting our QoS if the user does something wrong with their webhooks, and maybe later we can add some flag to the webhook saying that it is "disallow on failure", in case someone ever needs/prefers it like that.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,6 +473,77 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+		}
+
+		if resp.StatusCode/100 != 2 {
+			glog.Errorf("Error calling user.new webhook %s status code=%d body=%s", userWebhook.Url, resp.StatusCode, string(body))
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("false"))
+			return false
```
victorges: I think we should `return true` on every return call here, since that's only used to log the request or not. Since we make I/O and call external services here, might be interesting to always have the latency log.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -500,7 +500,7 @@ func (mc *mac) triggerUserNew(w http.ResponseWriter, r *http.Request, lines []st
 	}
 
 	if len(userWebhooks) == 0 {
-		glog.V(model.DEBUG).Infof("No user.new webhooks for user %s", userId)
+		glog.V(model.DEBUG).Infof("No playback.user.new webhooks for user %s", userId)
 		w.WriteHeader(http.StatusOK)
 		w.Write([]byte("false"))
```
victorges, line 543: Yeah this one should certainly be `"true"`, right?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -91,6 +93,15 @@ type (
 		startedAt          time.Time
 	}
 
+	userNewPayload struct {
+		StreamName   	   string   `json:"streamName"`
```
victorges, line 97: If we're only sending 1 information about the stream I believe it should be `streamId` instead.

We could also send "everything" like a full `stream` object, I only think it's complicated cause we'd have to filter any admin fields we might have received about it. So I think just the ID is good, and they can call the API for the full object if desired.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,14 +484,41 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+		OutputProtocol: lines[3],
+		RequestUrl:    	lines[4],
+		SessionID:     	lines[5],
+	}
+
+	b, err := json.Marshal(u)
+    if err != nil {
+        glog.Errorf("Error marshalling userNewPayload: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("false"))
```
victorges, line 516: I think we should also allow the request here IMO. An error marshalling means an internal error in our own logic to create the webhook payload, we should avoid breaking playback of all streams if that ever happens (which seems very unlikely anyway)

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,14 +484,41 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
 		return false
 	}
 
+	u := &userNewPayload{
+		StreamName:    	lines[0],
+		IpAddress:     	lines[1],
+		UserAgent:     	lines[2],
+		OutputProtocol: lines[3],
+		RequestUrl:    	lines[4],
+		SessionID:     	lines[5],
```
victorges, line 509: Hm, how come does `mist` have the stream name and session IDs nicely sent in the response like this? Are they Mist abstractions instead? If so, we should be interpreting all these and only send values that make sense to API users here (the "connecting" part of mist-api-connector 😆)

So for example, I already see that `lines[0]` is also used below as the playback ID. So it certainly does not mean the "Stream Name" to an API user. Nor do I think mist has any notion of our "session ID" in the API, that stays within `MistProcLivepeer` logic. 

So... You can only actually build this payload after you get the `*streamInfo` from the `pub2info` map. From it you can populate this payload with both some mist-specific data that make sense to API users like IP address, user agent, request URL, playback/stream ID, webhook ID, event ID etc.

victorges: Also, realized that the webhook payload should actually follow the format from our other webhooks, only changing what is inside the payload. This is where we create webhooks in the API: https://github.com/livepeer/livepeer-com/blob/6b86fb011c94e9c6dee5c9837110e0bdf7f08663/packages/api/src/webhooks/cannon.ts#L384-L392

So you should have all these fields: id, webhookId, createdAt, timestamp, event, stream (which I guess we can leave empty or only have an `{id: string}` field) and then a `payload` field where you can have the extra information specific to this webhook (you shouldn't need to repeat any info if it's already in the outer schema either, like the info about the stream)

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,14 +484,41 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
...
+
+	b, err := json.Marshal(u)
+    if err != nil {
+        glog.Errorf("Error marshalling userNewPayload: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("false"))
+        return false
+    }
+
+    payload := string(b)
```
victorges, line 520: **nit:** You can also keep it as `[]byte` and use `bytes.NewReader` or `bytes.NewBuffer` instead (not necessary)

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -505,38 +543,65 @@ func (mc *mac) triggerUserNew(w http.ResponseWriter, r *http.Request, lines []st
...
-		if userWebhook.Url == "" {
-			glog.Errorf("User webhook %s has no URL", userWebhook.ID)
-			continue
-		}
-		glog.V(model.DEBUG).Infof("Calling playback.user.new webhook %s", userWebhook.Url)
+		
+			if userWebhook.Url == "" {
+				glog.Errorf("User webhook %s has no URL", userWebhook.ID)
+				return true
+			}
```
victorges, line 552: I guess we can skip this check now that we are parsing the URL below?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -473,14 +484,41 @@ func (mc *mac) generateRouteKeys(stream *livepeer.CreateStreamResp) []string {
 	return keys
 }
 
+func (mc *mac) sign(sharedSecret string) string {
+	s := sha256.New()
+	s.Write([]byte(sharedSecret))
+	signature := hex.EncodeToString(s.Sum(nil))
+	return signature
```
victorges, line 491: Our signature is not just a SHA258, it's an HMAC of the message with the encryption key (hash of the encrypted payload?). This implementation must be compatible with our regular webhooks implementation, so we should change this to match it.

https://github.com/livepeer/livepeer-com/blob/6b86fb011c94e9c6dee5c9837110e0bdf7f08663/packages/api/src/controllers/helpers.ts#L19

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -505,38 +543,65 @@ func (mc *mac) triggerUserNew(w http.ResponseWriter, r *http.Request, lines []st
...
+			if err != nil {
+				glog.Errorf("Error parsing URL %s for user webhook %s err=%v", userWebhook.Url, userWebhook.ID, err)
+				return false
+			}
+
+			if userWebhook.SharedSecret != "" {
+				signature := mc.sign(userWebhook.SharedSecret)
+				timestamp := strconv.FormatInt(time.Now().Unix(), 10)
+				signature_header := fmt.Sprintf("t=%s,v1=%s", timestamp, signature)
+				req.Header.Add("Livepeer-Signature", signature_header)
```
victorges, line 573: This is not a correct implementation of the webhook signature algorithm.

You **must** include the payload of the request in the signed payload, otherwise the signature is reusable. The body **must** also contain the `timestamp`, as per the unofficial webhook payload schema (docs are the code lol) I mentioned. That guarantees that a timestamp change will change the body and then change the signature.

These implementations must be compatible so I suggest you check it here: https://github.com/livepeer/livepeer-com/blob/6b86fb011c94e9c6dee5c9837110e0bdf7f08663/packages/api/src/webhooks/cannon.ts#L397

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -505,38 +543,65 @@ func (mc *mac) triggerUserNew(w http.ResponseWriter, r *http.Request, lines []st
...
+			}
+
+			glog.V(model.DEBUG).Infof("Calling playback.user.new webhook %s", userWebhook.Url)
+
+			resp, err := http.DefaultClient.Do(req)
+			
+			if err != nil {
+				glog.Errorf("Error calling playback.user.new webhook %s err=%v", userWebhook.Url, err)
+				w.WriteHeader(http.StatusInternalServerError)
+				w.Write([]byte("false"))
```
victorges, line 583: Hmm so I see we return both a status code and write true/false on the body. What do these actually mean? 

I got that the boolean we return is only used to log or not, but the other ones are still a mystery to me.

internal/app/mistapiconnector/mistapiconnector_app.go
victorges: Nevermind, go does not return an error when you try to parse an empty string as a URL. I think this check might still be unnecessary tho, since the API is the one doing this kind of validation of the resources. We can assume they are valid if they exist, I guess.

internal/app/mistapiconnector/mistapiconnector_app.go
gioelecerati: Thank you, this and the hmac were a pretty bad overlook this morning, fixed it

gioelecerati: I temporarly removed the userAgent from the userNewPayload and added the timestamp. Will manage it better when I am onto which data we need to send to the user from Mist

internal/app/mistapiconnector/mistapiconnector_app.go
gioelecerati: In the USER_NEW payload `streamName` is not the the stream ID but actually the playback ID. I think `StreamID` here would make it confusing, as well as `PlaybackID` so I kept the Mist key name, wdyt?

victorges: I think webhook payloads are just like API endpoints. They are part of the public interface of the API. As such, they should use a consistent language between resources.

On this case, this StreamName is not something that makes sense for our current API abstractions. It would seem like it's the stream.name, but it's something else. For users of the API that something else also has an existing name, playbackId. So that's what I think we should would call it, even this didn't have to change for consistency with the other API webhooks anyway.

So you should just follow the schema that I linked in the other comment. You would just have a stream object here with all the fields of the stream, among some other default metadata fields.

There is also the payload field which is the freeform one, each webhook has its own. I think it would make sense to setup that webhook using the concepts of our current API, before a new one is redesigned. If you prefer to use mist terms tho, at least they will be isolated inside this payload field so it's likely less confusing.

iameli (changes requested):

iameli (changes requested):
Looks like this doesn't auto-create the USER_NEW trigger the way that the other Mist actions are auto-created. Could you figure out where that happens and adding it?

internal/app/mistapiconnector/mistapiconnector_app.go
iameli: Uncertain I agree; in the event that a user new trigger is configured, the expectation is that a stream is private. Once users start actually using this mechanism, even if there's a failure on our end, I think they'd expect "fail shut" behavior rather than "allow the entire world to watch the stream now".

internal/app/mistapiconnector/mistapiconnector_app.go
iameli: Disagree as per other point

internal/app/mistapiconnector/mistapiconnector_app.go
iameli: I gotta give this out today, so I'm going to strip the hook down to just this:

```
{
  "payload": {
    "requestUrl": "http://playback.livepeer.systems:8081/hls/d6b8nhc03ofgwx6v/index.m3u8?stream=d6b8nhc03ofgwx6v&proof=abc123"
  }
}
```

That'll presumably stay stable and contains all the information that they need to implement token-gating. We can add other fields and make more consistent based on what we want

iameli: Er, I guess we need timestamp as well for HMAC confirmation?

iameli:
Per conversation, let's do `--disable-trigger=USER_NEW` so we can short-circuit this in production for a while

iameli:
Seems like we need precaching here, per https://discord.com/channels/423160867534929930/974381655651385404/975155567054188655

apis/livepeer/livepeer.go
```diff
@@ -470,6 +477,12 @@ func (lapi *API) GetWebhooksForEvent(userId string, streamId string, event strin
 	return lapi.GetWebhooks(u)
 }
 
+// Update webhook status & response
+func (lapi *API) UpdateWebhookStatus(id string, status map[string]interface{}) bool {
+	go lapi.UpdateWebhook(id, status)
+	return true
```
victorges, line 483: This return value seems unnecessary. Actually the whole function. In go patterns you wouldn't need a function `g()` to do `g() { go f() }`, you just make `go f()` (and it's also clear on call-site that it is async). 

IOW The callers of `UpdateWebhookStatus` can call `go UpdateWebhook()`.

victorges: Side note, if you prefer to keep the 2 functions, might prefer to rename this helper to have a clearer name, since it's not really describing what it does in relation to the `UpdateWebhook` func. It could be called `UpdateWebhookAsync` or something like that, WDYT?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -579,34 +584,55 @@ func (mc *mac) triggerUserNew(w http.ResponseWriter, r *http.Request, lines []st
 			req.Header.Add("Livepeer-Signature", signature_header)
 		}
 
-		glog.V(model.DEBUG).Infof("Calling playback.user.new webhook %s", userWebhook.Url)
-
 		resp, err := http.DefaultClient.Do(req)
+		webhookStatus := livepeer.WebhookStatus{
+			LastTriggeredAt: time.Now().Unix(),
```
victorges, line 589: This will not output a timestamp compatible with the API. You need to do or have a helper like thiis:
https://github.com/livepeer/livepeer-data/blob/main/pkg/data/time.go#L16
to do a `UnixMillis`.

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -501,8 +501,14 @@ func (mc *mac) sign(body string, sharedSecret string) string {
 	return signature
 }
 
-func (mc *mac) triggerUserNew(w http.ResponseWriter, r *http.Request, lines []string, rawRequest string) bool {
+func (mc *mac) webhookStatusUpdate(status livepeer.WebhookStatus, id string) {
+	statusRequestUpdate := map[string]interface{}{
```
victorges, line 505: We should have proper request types for all these maps! Makes interfaces more maintainable and easier to use.

apis/livepeer/livepeer.go
gioelecerati: Yeah for now I prefer to keep the two functions, everything works fine and I want to ship it in the separate docker to debug bonfire, will get to this later down the road! Renamed to `async` and removed the boolean return btw

gioelecerati:
Consider closing this as the new access control is in production.
Need to check if this version is still used in any way

gioelecerati:
Closing it

PR #143, opened 2022-04-26
Add orchestrator summary log

leszko:
Adds a summary for the orch-tester job.

It logs a warning if more less than best 20 Os:
- either had success rate < 80%
- or round trip time over 1.0


iameli (approved):
I don't think I've ever looked through this code before, but LGTM.

Issue #146, opened 2022-05-02
record-tester: add e2e testing for NFT SDK


iameli:
* After record-tester finishes, wait for Asset to be available
* Run transcode task on that asset
* Wait for transcoded thing to become available
* Run export task
* Check that IPFS link is available

PR #148, opened 2022-05-17
Add E2E Testing for NFT SDK

figintern:
https://github.com/livepeer/stream-tester/issues/146

1. Refactor `record-tester` to use `go-api-client`

2. After `record-tester` finishes ->
* Wait for Asset to be available
* Run transcode task on that asset
* Wait for transcoding asset to complete
* Run export task on transcoded asset
* Wait for export to complete
* Check that IPFS link is available on export task

Testing locally:
```
~/livepeer/stream-tester % make
~/livepeer/stream-tester % ./build/recordtester -api-token 29b59cfa-3006-48b7-a5fd-fbd07073372e  -file https://storage.googleapis.com/lp_testharness_assets/bbb_sunflower_1080p_30fps_normal_2min.mp4 -test-dur 4m -continuous-test 15s -vod -stream-health true
...
I0517 04:09:47.721283   45384 recordtester_app.go:306] Record testing done, waiting for Asset to be available
I0517 04:09:47.721300   45384 recordtester_app.go:316] Waiting 30s for asset, elapsed=0s
I0517 04:10:18.172198   45384 recordtester_app.go:316] Waiting 30s for asset, elapsed=30s
I0517 04:10:48.666966   45384 recordtester_app.go:316] Waiting 30s for asset, elapsed=1m0s
I0517 04:11:19.198480   45384 recordtester_app.go:316] Waiting 30s for asset, elapsed=1m30s
I0517 04:11:50.226989   45384 recordtester_app.go:348] Asset is available id=56904bbd-46c2-4bc9-9d6a-874442340b22, transcoding taskId=cb594fd5-197c-4769-b1f8-27984623c66d outputAssetId=7b526a4e-4d28-40c7-a224-b8699c4695c9
I0517 04:11:50.227012   45384 recordtester_app.go:353] Waiting 15s for transcode output asset id=7b526a4e-4d28-40c7-a224-b8699c4695c9, elapsed=0s
I0517 04:12:06.143579   45384 recordtester_app.go:384] Transcode output asset id=7b526a4e-4d28-40c7-a224-b8699c4695c9 ready, exporting, taskId=1a51823c-3186-4330-a3a5-121335cacbad
I0517 04:12:06.143610   45384 recordtester_app.go:389] Waiting 5s for asset id=7b526a4e-4d28-40c7-a224-b8699c4695c9 to be exported, elapsed=0s
I0517 04:12:11.562321   45384 recordtester_app.go:403] Export success, task id=1a51823c-3186-4330-a3a5-121335cacbad ipfs link=https://ipfs.livepeer.com/ipfs/QmfPT1LxgpP1achTWECXkghTRqb3j7iifJSLDp4ergqSK2
I0517 04:12:11.562350   45384 recordtester_app.go:409] Done
```

go-api-client PR: https://github.com/livepeer/go-api-client/pull/7


internal/app/recordtester/recordtester_app.go
```diff
@@ -390,18 +456,12 @@ func (rt *recordTester) getIngestInfo() (*livepeer.Ingest, error) {
...
 	var err error
 	apiTry := 0
 	for {
-		session, err = rt.lapi.CreateStream(livepeer.CreateStreamReq{
-			Name:                streamName,
-			ParentID:            stream.ID,
-			Profiles:            standardProfiles,
-			Record:              true,
-			RecordObjectStoreId: rt.recordObjectStoreId,
-		})
```
victorges, line 404: My bad on creating this simpler `CreateStream` function only on `stream-tester` and not replicating the change on `go-api-client`. I believe it provides a better and simpler interface (instead of multiple functions with cryptic names like "ex" "ex2", etc).

So I'd say that `go-api-client` should be the one migrating to provide this single `CreateStream` function instead, not the other way around. WDYT?

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+	}
+
+	assetId := ""
+	wait_interval := 30 * time.Second
+	elapsed_time := 0 * time.Second
+asset:
+	for {
+		glog.Infof("Waiting %s for asset, elapsed=%s", wait_interval, elapsed_time)
+		time.Sleep(wait_interval)
+		elapsed_time = elapsed_time + wait_interval
```
victorges, line 318: Instead of accumulating only the time you sleep, it should be better to save the `starTime := time.Now()` and then log the elapsed time as `time.Since(startTime)`

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+		}
+
+		assets, err := rt.lapi.ListAssets() // assets are retrieved in order of most recent first
+		if err != nil {
+			glog.Errorf("Error listing assets err=%v", err)
+			return 241, err
+			// exit(241, fileName, *fileArg, err)
+		}
+		for _, asset := range *assets { // TODO: only need to check first n assets due to ordering
+			if asset.Name == fmt.Sprintf("live-to-vod-%s", sess.ID) {
```
victorges, line 330: I think it will be even better if we start updating the `session` object with the ID of the recording asset instead. That update should be done [from here](https://github.com/livepeer/livepeer-com/blob/e5b0ec26ca996ec6cb997558bc41b25ecce7239a/packages/api/src/webhooks/cannon.ts#L516) and should be really straightforward. In a simplified way, probably a single line change as well like this:
```ts
await this.db.stream.update(sessionId, { recordingAssetId: asset.id });
```
(we'd still need to add the field to the schema, and might also wanna do that update from the outer function instead of that line I linked)

Then here instead of listing all assets, we keep `GET`ing the session until it comes with that asset ID, then we can continue from there.

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+				assetId = asset.ID
+				break asset
+			}
+		}
+
+		// TODO: timeout required here?
+	}
+
+	task_name := fmt.Sprintf("live-to-vod-transcode-%s", sess.ID)
```
victorges, line 339: This is the asset name, not the task name. The task name is not configurable. Related to the comments I made that the function below should be called `TranscodeAsset`, not `CreateTranscodeTask`.

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+	createTranscodeTaskResp, _ := rt.lapi.CreateTranscodeTask(assetId, task_name, livepeerAPI.StandardProfiles[0])
+	if err != nil {
```
victorges, line 341: You're checking the error here but you're ignoring it in the line above. 
```suggestion
	createTranscodeTaskResp, err := rt.lapi.CreateTranscodeTask(assetId, task_name, livepeerAPI.StandardProfiles[0])
	if err != nil {
```

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+	transcodeTask := createTranscodeTaskResp.Task
+	transcodeAsset := createTranscodeTaskResp.Asset
+	glog.Infof("Asset is available id=%s, transcoding taskId=%s outputAssetId=%s", assetId, transcodeTask.ID, transcodeAsset.ID)
+
+	wait_interval = 15 * time.Second
+	elapsed_time = 0 * time.Second
+	for {
+		glog.Infof("Waiting %s for transcode output asset id=%s, elapsed=%s", wait_interval, transcodeAsset.ID, elapsed_time)
+		time.Sleep(wait_interval)
+		elapsed_time = elapsed_time + wait_interval
```
victorges, line 355: Same regarding elapsed time.

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+		glog.Errorf("Error creating export task err=%v", err)
+		return 245, err
+		// exit(245, fileName, *fileArg, err)
+	}
+	exportTask := createExportTaskResp.Task
+	glog.Infof("Transcode output asset id=%s ready, exporting, taskId=%s", transcodeAsset.ID, exportTask.ID)
+
+	wait_interval = 5 * time.Second
+	elapsed_time = 0 * time.Second
+	for {
```
victorges, line 389: For these wait intervals, you should also consider using a `time.Ticker` instead for go-ishness. e.g.
```suggestion
	ticker := time.NewTicker(5 * time.Second)
	for range ticker {
	   ...
```

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,111 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+		glog.Infof("Waiting %s for asset id=%s to be exported, elapsed=%s", wait_interval, transcodeAsset.ID, elapsed_time)
+		time.Sleep(wait_interval)
+		elapsed_time = elapsed_time + wait_interval
+		if err = rt.isCancelled(); err != nil {
+			return 0, err
+		}
+
+		task, err := rt.lapi.GetTask(exportTask.ID)
```
victorges, line 397: We should probably use the same mechanism to wait for both the asset transcode task as well as the asset export task. This means that in both cases we can just wait for the task and get the asset in the end of the process, instead of polling for the task as you did on the transcode case.

You could use this as an example and create a similar helper here that you can re-use: https://github.com/livepeer/video-nft/blob/main/src/minter.ts#L480

With a separate helper that only waits for a task (which can be in the `go-api-client` lib as well sine it's a pretty common logic), the logic here should become a little cleaner as well.

victorges (commented):
The overall logic here looks good! The only bigger issue is that, even though I think I suggested that myself, I'm actually having second thoughts about including all these tests in the `record-tester` flow. I believe most if not all of the logic here will be the same anyway, the only difference is that I think we should have a separate command/tester just for that.

The main reasons for that are:
 - Including these tests in the same flow will introduce a big pause between test streams. There's a whole minute until a recording _starts_ being processed to become an asset, so I expect that to add at least 5 minutes before new livestreams are made. This will increase our time for getting an alert when a region goes down.
 - Our recordings (and probably transcoding) output is not deterministic. This means that by exporting the transcoded recording to IPFS, we are going to create a new file every time. Piñata storage is preeetty expensive (I think 5x the price of an S3), so I think this bill could grow pretty quickly. We already pay thousands of dollars a month for old test recordings saved on Google S3, on Piñata it's gonna 📈. We could either find some way to unpin the file after the test is done, but I think it will be simpler if we just always export the same file to IPFS.

So a simpler VOD (NFT) specific tester could work like:
 - Start from the same video file as record-tester
 - Import that to an asset 1
 - Try transcoding that to another asset 2
 - Try exporting the asset 1 (not asset 2) to IPFS
 - Make sure all that works
 - Success, sleep and try again in a few (just like `continuous_record_tester` here)

Let's talk more about this next week! I'll bring it up on our planning and we can talk about it ad-hoc or schedule additional meetings to go through it as well.

internal/app/recordtester/recordtester_app.go
```diff
@@ -155,14 +121,9 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
-	var stream *livepeer.CreateStreamResp
+	var stream *api.CreateStreamResp
 	for {
-		stream, err = rt.lapi.CreateStream(livepeer.CreateStreamReq{
-			Name:                streamName,
-			Profiles:            standardProfiles,
-			Record:              true,
-			RecordObjectStoreId: rt.recordObjectStoreId,
-		})
+		stream, err = rt.lapi.CreateStreamEx2(streamName, true, rt.recordObjectStoreId, "", nil, api.StandardProfiles...)
```
victorges, line 126: OK I saw this and decided to pay my debts on the `go-api-client` repo by properly upstreaming the `CreateStream` changes there: https://github.com/livepeer/go-api-client/pull/10

If you're OK with it, mind reviewing that one so I can merge and here you can use the same `CreateStream` function that was used before?

internal/app/recordtester/recordtester_app.go
```diff
@@ -341,6 +302,107 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 	// <-downloader.Done()
 	// glog.Infof(`Pulling stopped after %s`, time.Since(started))
 	// exit(55, fileName, *fileArg, err)
+
+	glog.Info("Record testing done, waiting for Asset to be available")
```
victorges, line 306: Did you forget to delete these changes?

go.mod
```diff
@@ -34,6 +34,8 @@ require (
 	google.golang.org/grpc v1.38.0
 )
 
+require github.com/livepeer/go-api-client v0.1.6-0.20220613100107-b8ace0aeee99
```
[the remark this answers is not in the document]
hjpotter92: just running `go mod tidy` should be enough. :)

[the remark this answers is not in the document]
hjpotter92: then this split require is the recommended method from golang

i remember the split requires happening after `go mod tidy` for direct/indirect dependencies for some other project recently.

cmd/recordtester/recordtester.go
```diff
@@ -167,8 +167,12 @@ func main() {
...
 	}
 
-	lapi = livepeer.NewLivepeer2(*apiToken, *apiServer, nil, 8*time.Second)
-	lapi.Init()
+	lApiOpts := api.ClientOptions{
+		Server:      *apiServer,
+		AccessToken: *apiToken,
+		Timeout:     8 * time.Second,
+	}
+	lapi, _ = api.NewAPIClientGeolocated(lApiOpts)
```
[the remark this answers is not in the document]
victorges: Ah, right! I hadn't checked the implementation, just assumed that the second value here was an `error`. The confusion is caused by an anti-pattern I created myself there, of calling the `MustX` function from inside a function that is not called `MustX` itself 😆 

Maybe it should have been `MustNewAPIClientGeolocated` just to leave it clear that it crashes on any error.

But don't bother about this! This looks good and we don't need to worry about refactoring `go-api-client` for this.

internal/app/vodtester/vodtester_app.go
```diff
@@ -0,0 +1,161 @@
...
+	vt.cancel()
+}
+
+func (vt *vodTester) Done() <-chan struct{} {
+	return vt.ctx.Done()
+}
+
+func (rt *vodTester) Clean() {
+	// TODO
+}
```
victorges, line 161: Instead of having this clean function, you could simply do a `defer api.DeleteAsset(asset.ID)` after every asset you created on the test routine above.

We probably don't want to infinitely stack up the account with so many assets from these tests. Although we won't be really deleting the files from cloud storage nor even the objects from DB anyway 🤷
So it's more not to pollute the API responses/dashboard than for any concrete resource optimization.

internal/app/vodtester/vodtester_app.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx, cancel := context.WithCancel(gctx)
+	vt := &vodTester{
+		lapi:   opts.API,
+		ctx:    ctx,
+		cancel: cancel,
+	}
+	return vt
+}
+
+func (vt *vodTester) Start(fileUrl string, testDuration, pauseDuration time.Duration) (int, error) {
```
victorges, line 43: Both `testDuration` and `pauseDuration` are actually unused in `vodTester`. Should remove the args.

internal/app/vodtester/continuous_vod_tester.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return cvt
+}
+
+func (cvt *continuousVodTester) Start(fileName string, testDuration, pauseDuration, pauseBetweenTests time.Duration) error {
+	messenger.SendMessage(fmt.Sprintf("Starting continuous vod test of %s", cvt.host))
+	try := 0
+	notRtmpTry := 0
+	maxTestDuration := 4*testDuration + pauseDuration + 15*time.Minute
+	for {
```
victorges, line 67: WDYT of enforcing a test cadence here, just so we don't do tests as quickly as possible on the VOD API? Say soon we might optimize the transcode/prepare tasks and then this tests would start flying really fast.

So here, we can actually use this `testDuration` which already exists and is passed around by the `record-tester` command to force a minimum delay between the tests. You can enforce it easily here with a ticker!
```suggestion
func (cvt *continuousVodTester) Start(fileName string, testDuration, pauseDuration, pauseBetweenTests time.Duration) error {
	messenger.SendMessage(fmt.Sprintf("Starting continuous vod test of %s", cvt.host))
	try := 0
	notRtmpTry := 0
	maxTestDuration := 4*testDuration + pauseDuration + 15*time.Minute
	ticker := time.NewTicker(testDuration)
	defer ticker.Stop()
	for range ticker.C {
```
And so the loop will only be "triggered" every `testDuration` 😌 

WDYT?

internal/app/vodtester/continuous_vod_tester.go
```diff
@@ -0,0 +1,177 @@
...
+		vtOpts:                  opts.VodTesterOptions,
+	}
+	return cvt
+}
+
+func (cvt *continuousVodTester) Start(fileName string, testDuration, pauseDuration, pauseBetweenTests time.Duration) error {
+	messenger.SendMessage(fmt.Sprintf("Starting continuous vod test of %s", cvt.host))
+	try := 0
+	notRtmpTry := 0
+	maxTestDuration := 4*testDuration + pauseDuration + 15*time.Minute
```
victorges, line 66: This is not valid for the VOD test, it was calculated for the livestram test which does 2 streams.

This should just be `testDuration` IMO (and so we don't need the intermediate variable), which is already a high value like `4m` right now.

victorges (approved):
LGTM!

PR #152, opened 2022-05-24
Add orchestrator_uri tag while querying Prometheus for avgRoundTripTime

### LINKED ISSUE #151: Fix Prometheus queries after adding Per Orchestrator metrics
In https://github.com/livepeer/go-livepeer/issues/2371 we tagged 3 metrics with the orchestrator address and service URI labels. This change may not be backwards-compatible for some queries we use. In particular, it's an issue for this query we use to get average round trip time:

```
rate(livepeer_transcode_overall_latency_seconds_sum[1m])/rate(livepeer_transcode_overall_latency_seconds_count[1m])
```

Since we get always the first entry in from the query, it resulted in the bug that not every stream has latency equal to zero.

leszko:
fix #151


red-0ne (approved):
LGTM

PR #161, opened 2022-06-01
Embed streamtester server in orchtester job

### LINKED ISSUE #155: Embed Streamtester into Orchtester
Orchtester uses Streamtester for sending test streams to orchestrators. Currently, we deploy Orchtester and Streamtester as two separate entities ([orchtester](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/templates/orch-tester.yaml), [streamtester](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/templates/stream-tester.yaml)) and specify the dependency between them at the HTTP level, with the [`-streamtester` flag](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L48). This approach is complex and hard to run locally.

We should embed Streamtester into Orchtester, so that they run in one process. We can do it with the following approach:
1. Check if the flag `-streamtester` is specified; if not, then use "embedded streamtester"
2. Start "embedded streamtester" with the following code

```go
s := server.NewStreamerServer(false, "", "", 4242)
go func() {
	s.StartWebServer(context.TODO(), "127.0.0.1:7934")
}()
```
3. Use `127.0.0.1:7934` endpoint for the orchtester -> streamtester communication.

**Additional comments**
- Please find the PoC for this work [here](https://github.com/livepeer/stream-tester/commit/c87ce0ddfba6157bc130ce1e56d31f9854b7f0c4)
- Some refactoring of [`streamtester.main()`](https://github.com/livepeer/stream-tester/blob/master/cmd/streamtester/streamtester.go#L37) may be required if we want to pass streamtester execution parameters

leszko:
Changes:
- Embed streamtester into orchtester
- Minor refactor (`defaultAddr()`)

fix #155 

Tested in our staging environment and it works correct. Here's the [infra part](https://github.com/livepeer/livepeer-infra/compare/master...rafal/155-embed-streamtester) I used to deploy the change into staging.

### DIFF
--- a/cmd/orch-tester/orch_tester.go
+++ b/cmd/orch-tester/orch_tester.go
@@ -10,12 +10,14 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/livepeer/stream-tester/internal/server"
 	"io/ioutil"
 	"log"
 	"math"
 	"math/rand"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -23,6 +25,7 @@ import (
 	"github.com/golang/glog"
 	apiModels "github.com/livepeer/leaderboard-serverless/models"
 	"github.com/livepeer/m3u8"
+	streamtesterMetrics "github.com/livepeer/stream-tester/internal/metrics"
 	"github.com/livepeer/stream-tester/internal/testers"
 	"github.com/livepeer/stream-tester/model"
 	streamerModel "github.com/livepeer/stream-tester/model"
@@ -33,6 +36,8 @@ import (
 
 const defaultHost = "127.0.0.1"
 const streamTesterPort = "7934"
+const streamTesterLapiToken = ""
+const streamTesterMistCreds = ""
 const prometheusPort = "9090"
 const bcastMediaPort = "8935"
 const bcastRTMPPort = "1935"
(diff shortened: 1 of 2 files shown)


cmd/orch-tester/orch_tester.go
```diff
@@ -71,30 +74,26 @@ func main() {
...
-	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
-	subgraphURL, err := defaultAddr(*subgraph, "127.0.0.1", "8080")
-	if err != nil {
-		log.Fatal(err)
+	if *streamTester == "" {
+		// start streamtester service
+		glog.Info("Using embedded broadcaster")
```
victorges, line 82: typo on broadcaster + log can be the comment
```suggestion
		glog.Info("Starting embedded streamtester service")
```

cmd/orch-tester/orch_tester.go
```diff
@@ -71,30 +74,26 @@ func main() {
...
+		hostName, _ := os.Hostname()
+		streamtesterMetrics.InitCensus(hostName, model.Version, "streamtester")
+		s := server.NewStreamerServer(false, "", "", 4242)
```
victorges, line 85: Might want to provide `*mistPort` customizability through a CLI flag as well. If you believe it's unnecessary, WDYT of creating a constant for it just so it's clear what it means here? I initially looked at this and was confused thinking this was the port that the stream server would listen too. Especially confusing since there's another port below.

Same for `*mistCreds` and Livepeer `*apiToken`. Or are all of those unused in the use that we make of the strema-tester from the orch-test? (then it's a `StreamerServer` interface issue, no need to bother refactoring).

victorges (approved):
LGTM! Only some nits

cmd/orch-tester/orch_tester.go
leszko: I think `mistCreds` and `apiToken` are not needed (at least I don't see why we could need them. 

Totally, agree with constants, updated in 254c84ad48709ef8f3f622d6ae0cc66b4dd1c702

red-0ne (approved):
LGTM!

Issue #159, opened 2022-06-01
Deploy Orchtester without dependencies


leszko:
When Orchtester is finally the self-container process, we need to update `livepeer/livepeer-infra`, so that it starts Orchtester without any dependencies.

Required changes:
1. Remove the [orchtester helm chart dependencies (broadcaster, monitoring)](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/Chart.yaml#L23)
2. Remove [stream-tester.yaml](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/templates/stream-tester.yaml)
3. Update [values.yaml](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/values.yaml)
4. Update all values files for orchtester (https://github.com/livepeer/livepeer-infra/tree/master/values)

After the changes I suggest the following roll out actions:
1. Deploy to staging => check if everything works fine
2. Deploy to prod in one region => check if everything works fine
3. Deploy to prod half of the regions => check if everything works fine
4. Deploy to prod all regions => check if everything works fine

For the whole context of the change, please check [Test Streams Architecture Design Doc](https://www.notion.so/livepeer/Test-Streams-Architecture-932fe3a75c874a36840fa85d4273b6db)

PR #162, opened 2022-06-06
Embed broadcaster in orchtester job

### LINKED ISSUE #156: Embed Broadcaster into Orchtester
Orchtester uses Broadcaster for sending test streams to orchestrators. Technically, Orchtester uses Streamtester which uses Broadcaster. Currently, we deploy Orchtester and Broadcaster as two separate entities ([orchtester](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/templates/orch-tester.yaml), [broadcaster](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/Chart.yaml#L24)) and specify the dependency between them at the HTTP level, with the [`-broadcaster` flag](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L49). This approach is complex and hard to run locally.

We should embed Broadcaster into Orchtester, so that they run in one process. We can do it with the following approach:
1. Check if the flag `-broadcaster` is specified; if not, then use "embedded broadcaster"
2. Start "embedded broadcaster" with the following code

```go
cfg := starter.DefaultLivepeerConfig()
cfg.Broadcaster = boolPointer(true)
go func() {
	starter.StartLivepeer(context.TODO(), cfg)
}()
```
3. Use `127.0.0.1:8935` endpoint for the broadcaster communication.
4. Set up [broadcaster timeout configuration](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L208) (currently set with the `LP_IS_ORCH_TESTER` env variable)

**Additional comments**
- Please find the PoC for this work [here](https://github.com/livepeer/stream-tester/commit/f34bc96574a88d1f03b0b067d87c2c385867dfc9)
- We will need to a dependency from `livepeer/stream-tester` to `livepeer/go-livepeer` in [go.mod](https://github.com/livepeer/stream-tester/blob/master/go.mod).
- Similar approach is already used in go-livepeer [E2E Tests](https://github.com/livepeer/stream-tester/blob/master/go.mod).

### LINKED ISSUE #158: Build a separate Orchtester Docker image
Currently, Orchtester does not have its own Docker image. It's run from the [`livepeer/streamtester`](https://hub.docker.com/r/livepeer/streamtester) image by specifying a [different CMD in the deployment configuration](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/values.yaml#L44).

We should provide a separate `livepeer/orchtester` Docker image built in the GH Actions.

leszko:
Changes:
- If the broadcaster flag not specified, then start embedded broadcaster service
- Add go-livepeer dependency into `go.mod`
- Create a separate Dockerfile `Dockerfile.orch-tester` to build orchtester into `livepeer/orch-tester` image
- Create related GH Action workflow
- Remove `orch-tester` from the `livepeer/streamtester` Docker image (plus remove unsed code)
- Add parsing flags from config file and env variables (it'll simplify the infra configuration)
- Minor refactor of `orch_tester.go`

fix #156 
fix #158 

Tested in our staging environment and it works correct. Here's the [infra part](https://github.com/livepeer/livepeer-infra/compare/master...rafal/156-embed-broadcaster) I used to deploy the change into staging.


Dockerfile.orch-tester
```diff
@@ -0,0 +1,35 @@
...
+COPY . .
+
+RUN go build -ldflags="-X 'github.com/livepeer/stream-tester/model.Version=$version' -X 'github.com/livepeer/stream-tester/model.IProduction=true'" -tags mainnet cmd/orch-tester/orch_tester.go
+
+FROM debian:stretch-slim
+
+RUN apt update \
+  && apt install -y ffmpeg ca-certificates \
+  && apt clean && apt autoclean
+RUN ffmpeg -version
```
victorges, line 25: Did you copy this bit from the `task-runner`? If so, I believe we don't need it here! And removing it will make the docker image build much faster :)

We actually only have that in the `task-runner` because not only it has a compile-time dependency on go-livepeer which requires the compiled ffmpeg bins [1], but it also calls `ffmpeg` (actually `ffprobe`) as a separate process on runtime. So we need to install it like this as well.

[1] The `task-runer` actually doesn't even use `go-livepeer` code that uses `ffmpeg`, only `drivers` lib. But there is some internal dependency leak on `go-livepeer` through the `common` package and so every package in `go-livepeer` depends on the compiled `ffmpeg` 🥲

victorges: Actually just saw that there is an installation like this on the regular `stream-tester` docker image as well. Not sure if it is needed by the `orch-tester` binary specifically tho 🤔 

Worst case it's not a regression on build time anyway, so no worries keeping this anyway to stay on the safe side.

.github/workflows/orch-tester-dockerimage.yml
```diff
@@ -0,0 +1,29 @@
...
+    - uses: actions/checkout@v1
+    - name: Docker Registry
+      env:
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+      run: docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+
+    - name: Tags
+      id: tags
+      uses: livepeer/action-gh-release-tags@v0
```
victorges, line 18: Some of the other workflows have this:
```yaml
      with:
        always-latest-on-branch: master
```

Do you think it'd be worth it here as well? Otherwise, it will only be tagged `latest` on dockerhub when you explicitly create a semver tag with a stable version (that might be desired as well)

Dockerfile.orch-tester
```diff
@@ -0,0 +1,35 @@
...
+RUN update-ca-certificates
+
+WORKDIR /root
+ENV PKG_CONFIG_PATH /root/compiled/lib/pkgconfig
+
+COPY --from=builder /root/compiled /root/compiled/
+COPY --from=builder /root/official_test_source_2s_keys_24pfs_30s_hls official_test_source_2s_keys_24pfs_30s_hls
+COPY --from=builder /root/orch_tester orch_tester
+
+CMD /root/orch_tester -v 6
```
victorges, line 35: Consider making this an `ENTRYPOINT` + `CMD` instead:
```suggestion
ENTRYPOINT ["/root/orch_tester"]
CMD ["-v", "6"]
```

This will allow the `orch_tester` to be run with custom arguments more easily, which will make it easier for Os to use it locally as well (e.g. `docker exec livepeer/orch-tester -- my custom args` instead of having to repeat the `/root/orch_tester` bit)

Also, maybe the `-v 6` bit should be specified as an environment variable instead like `ENV OT_V 6`, so its more easily overridable through the env but not automatically overridden by custom commands. That would be:
```suggestion
ENV OT_V 6
ENTRYPOINT ["/root/orch_tester"]
```
This also guarantees that setting a custom `OT_V` env still works, since I believe CLI args have precedence over env in our flag parsing lib 🤔

cmd/orch-tester/orch_tester.go
```diff
@@ -70,7 +93,22 @@ func main() {
...
+	bCfg.datadir = flag.String("datadir", "", "Directory that data is stored in")
+	bCfg.ethPassword = flag.String("ethPassword", "", "Password for existing Eth account address")
+	bCfg.maxTicketEV = flag.String("maxTicketEV", "3000000000000", "The maximum acceptable expected value for PM tickets")
+	bCfg.maxPricePerUnit = flag.Int("maxPricePerUnit", 0, "The maximum transcoding price (in wei) per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price")
+
+	// Config file
+	_ = flag.String("config", "", "Config file in the format 'key value', flags and env vars take precedence over the config file")
+	err := ff.Parse(flag.CommandLine, os.Args[1:],
+		ff.WithConfigFileFlag("config"),
+		ff.WithEnvVarPrefix("OT"),
```
victorges, line 109: Remember to update our deployments in `livepeer-infra` to include this prefix in env vars as well

cmd/orch-tester/orch_tester.go
```diff
@@ -266,6 +299,68 @@ func defaultAddr(addr, defaultHost, defaultPort string) string {
...
+	go func() {
+		addr := fmt.Sprintf("%s:%s", "0.0.0.0", streamTesterPort)
+		s.StartWebServer(ctx, addr)
+	}()
+}
+
+func startEmbeddedBroadcaster(ctx context.Context, bCfg broadcasterConfig, presets string) {
+	glog.Info("Starting embedded broadcaster service")
+
+	// Increase Broadcaster timeouts
```
victorges, line 316: Nice 😄

cmd/orch-tester/orch_tester.go
```diff
@@ -80,14 +118,15 @@ func main() {
...
-		s := server.NewStreamerServer(false, streamTesterLapiToken, streamTesterMistCreds, 4242)
-		go func() {
-			addr := fmt.Sprintf("%s:%s", "0.0.0.0", streamTesterPort)
-			s.StartWebServer(ctx, addr)
-		}()
+		startEmbeddedStreamTester(ctx)
+	}
+
+	var bcastHost string
+	if *broadcaster == "" {
```
victorges, line 125: Is it possible to use only the embedded stream-tester or broadcaster but not both? It seems like at least the embedded broadcaster is configured explicitly with the local stream tester (e.g. using `defaultHost` for the orch webhook). I think that's completely fine, but maybe we should have a check here to make sure that either none or both of these fields are set, failing the `orch_tester` with an error otherwise.

cmd/orch-tester/orch_tester.go
```diff
@@ -266,6 +299,68 @@ func defaultAddr(addr, defaultHost, defaultPort string) string {
...
+	cfg.LocalVerify = boolPointer(false)
+	cfg.HttpIngest = boolPointer(true)
+	cfg.TranscodingOptions = &presets
+	cfg.MaxTicketEV = bCfg.maxTicketEV
+	cfg.MaxPricePerUnit = bCfg.maxPricePerUnit
+	cfg.CliAddr = stringPointer(fmt.Sprintf("0.0.0.0:%s", bcastCliPort))
+	cfg.Broadcaster = boolPointer(true)
+	go func() {
+		starter.StartLivepeer(ctx, cfg)
+	}()
```
victorges, line 343: **nitnit:** just style
```suggestion
	go starter.StartLivepeer(ctx, cfg)
```

cmd/orch-tester/orch_tester.go
```diff
@@ -266,6 +299,68 @@ func defaultAddr(addr, defaultHost, defaultPort string) string {
...
+	rCtx, _ := context.WithTimeout(ctx, bcastReadyTimeout)
+	statusEndpoint := fmt.Sprintf("http://%s:%s/status", bcastHost, bcastCliPort)
+	ticker := time.NewTicker(200 * time.Millisecond)
+	for {
+		select {
+		case <-rCtx.Done():
+			glog.Error("Waiting for broadcaster timed out")
+			return
+		case <-ticker.C:
+			resp, err := http.Get(statusEndpoint)
```
victorges, line 356: Maybe make sure to call `resp.Body.Close()` when you do get a response here (not really sure if its necessary, but maybe not calling it can cause a resource leak)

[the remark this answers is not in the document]
victorges: I believe the `ticker` will just accumulate one other tick and we will receive immediately here on the next iteration. It won't accumulate multiple in-flight requests tho.

Either way, what is the usual time that the B takes to become online? Maybe it will be better to set a higher check delay here, like at least `1*time.Second`, to avoid firing too many requests. WDYT?

internal/metrics/census.go
```diff
@@ -118,7 +118,7 @@ func InitCensus(nodeID, version, namespace string) {
 	baseTags := []tag.Key{Census.kNodeID}
 	views := []*view.View{
 		{
-			Name:        "versions",
+			Name:        "streamtester_versions",
```
victorges, line 121: Will this have any relevant effect the other binaries built on this repository?

victorges (approved):
LGTM!

Dockerfile.orch-tester
victorges: Oh, maybe for the `-v` flag to work as an env var you will need to do sth like this? 
 - https://github.com/livepeer/task-runner/blob/71f66c7621439b3eb0350f63f2c5db7ea594217b/cmd/task-runner/task-runner.go#L66-L67 
 - https://github.com/livepeer/task-runner/blob/71f66c7621439b3eb0350f63f2c5db7ea594217b/cmd/task-runner/task-runner.go#L77

(and if you do that, maybe we don't even need this default value in the Dockerfile. We can just set the default value of that custom flag to be `6`)

.github/workflows/orch-tester-dockerimage.yml
leszko: Yeah, good idea. Added in 64379c7c3d09e000596f0eee6dc69816c9239596

I see `mist-api-connector-dockerimage.yml` don't have `always-latest-on-branch: master`. Do you think it might be worth to add it there as well? @victorges

Dockerfile.orch-tester
leszko: Right, FFMPEG is not need at runtime, updated in 8a09f0ee0b69779386991027e8ab3eae25e9a1a6

Dockerfile.orch-tester
leszko: Right, update `-v` flag to the same way we use it in `go-livepeer` (a937c5a5b4949c50ef934de551735c336d8b26f8)

leszko: Updated ENTRYPOINT in e2ac4b42571044685e56043314e0c32d852dd9eb

cmd/orch-tester/orch_tester.go
leszko: Yes, I'll update the whole `livepeer-infra` as part of this GH Issue https://github.com/livepeer/stream-tester/issues/159

cmd/orch-tester/orch_tester.go
leszko: Good point. updated in ee607027a515b64263b6972cf2eb839f1df9563f

cmd/orch-tester/orch_tester.go
leszko: 😄

cmd/orch-tester/orch_tester.go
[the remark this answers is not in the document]
leszko: Yeah, as mentioned by Victor, there should not be any problem of ticker running twice.

Updated to `1s` in 831d7cf894bc2bc73c27645b03be380897a5a298

leszko: Updated in cb55397a5b5d00c8754f8e15403454470013ddaa

internal/metrics/census.go
leszko: Yes, but I think we don't use this metric anywhere.

red-0ne (approved):
LGTM

PR #163, opened 2022-06-09
Use embedded broadcaster metrics instead of a separate monitoring service

### LINKED ISSUE #138: Fix Upload Time in Stream Tester
The upload time is not logged in the Stream Tester data. It's always `0.0`.

For example, https://interptr-latest-test-streams.vercel.app/?address=0x9d5611bf0dadddb4441a709141d9229d7f6b3e47, always shows only `0.0` in the `Upload Time` column.

I believe the bug is related to the [Per-stream metrics PR](https://github.com/livepeer/go-livepeer/pull/2238), which adds the transcoder URI to the metrics.

### LINKED ISSUE #157: Remove Monitoring from Orchtester's dependencies
Orchtester uses Monitoring for fetching metrics exported by Broadcaster while sending test streams to orchestrators. Monitoring is deployed as a separate entity [while deploying Orchtester](https://github.com/livepeer/livepeer-infra/blob/master/charts/orch-tester/Chart.yaml#L28). It consumes a lot of resources and we use it only for fetching a few metrics:
- [`livepeer_segment_source_upload_failed_total`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L445)
- [`livepeer_discovery_errors_total`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L462)
- [`livepeer_segment_transcode_failed_total`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L480)
- [`livepeer_download_time_seconds_sum`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L407)
- [`livepeer_transcode_overall_latency_seconds_sum`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L424)
- [`livepeer_source_segment_duration_seconds_sum`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L373)
- [`livepeer_upload_time_seconds_sum`](https://github.com/livepeer/stream-tester/blob/master/cmd/orch-tester/orch_tester.go#L390)

I believe we could replace replace this whole Prometheus with a local custom exporter attached to Broadcaster in the following manner:
```go
type OrchMetricsExporter struct {
}

func (e OrchMetricsExporter) ExportView(viewData *view.Data) {
	// simple logic to record these 3-4 metrics we're interested in
}
...
orchMetrics := OrchMetricsExporter{}
view.RegisterExporter(orchMetrics)
```

Then, there will be no need to deploy Monitoring at all.

I suggest the following approach:
1. Check if the flags `-broadcaster` and `-metrics` are not specified; if not then
2. Attach custom local exporter
3. Use data from the custom local exporter instead of the Prometheus queries sent to Monitoring

**Additional comments**
- Please find the PoC for this work [here](https://github.com/livepeer/stream-tester/commit/54c60453741a8de882cc36a4ed5357dff21e6b7f)

leszko:
Changes:
- If the `metrics` flag is not specified, then user the embedded metrics exporter
- Refactoring of Prometheus queries
- Additionally, if fixes some long-lasting bug #138 

fix #157 
fix #138 

Tested in our staging environment and it works correct. Here's the [infra part](https://github.com/livepeer/livepeer-infra/compare/master...rafal/157-embed-metrics) I used to deploy the change into staging.


cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
 	maxPricePerUnit *int
 }
 
+type broadcasterMetrics struct {
```
victorges, line 73: WDYT of moving these `broadcasterMetrics` to a separate file? Could even be in this same directory not to overcomplicate, but just to keep things separate

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+	}
+}
+
+func (bm *broadcasterMetrics) reset() {
+	bm.mu.Lock()
+	defer bm.mu.Unlock()
+
+	bm.rateSum = map[string]float64{}
+	bm.rateCount = map[string]int{}
+	bm.incErrs = map[string]int{}
```
victorges, line 105: Don't you need to reset the other fields here as well?

victorges: Also, if we do reset everything here like this, we could even make the constructor call this with something like
```
bm := &broadcasterMetrics{}
bm.reset()
return bm
```
WDYT?

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+	bm.handleErrorCountMetrics(rows, name)
+}
+
+func (bm broadcasterMetrics) handleDistributionMetrics(rows []*view.Row, name string) {
+	supportedMetrics := map[string]bool{
+		"source_segment_duration_seconds":   true,
+		"transcode_overall_latency_seconds": true,
+		"upload_time_seconds":               true,
+		"download_time_seconds":             true,
+	}
```
victorges, line 128: Could this be a global/static `var`?

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+
+	// Get the rate between the last measurement
+	rateSum := sum - bm.lastSum[name]
+	rateCount := count - bm.lastCount[name]
+
+	bm.rateSum[name] += rateSum
+	bm.rateCount[name] += rateCount
+	bm.lastSum[name] = sum
+	bm.lastCount[name] = count
```
victorges, line 151: I'm wondering if we really need the `lastSum/Count` values here, I might not have picked up some of the logic. Would this generate the same results?
```suggestion
	bm.rateSum[name] = sum
	bm.rateCount[name] = count
```

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+}
+
+func (bm broadcasterMetrics) handleErrorCountMetrics(rows []*view.Row, name string) {
+	supportedMetrics := map[string]bool{
```
victorges, line 155: Same for static var

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+		}
+		if errCode == "" {
+			return
+		}
+
+		d, ok := r.Data.(*view.CountData)
+		if ok {
+			inc := int(d.Value) - bm.lastErrs[errCode]
+			bm.incErrs[errCode] += inc
+			bm.lastErrs[errCode] = int(d.Value)
```
victorges, line 179: Same for maybe not needing the value of the last observation.

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+		return 0
+	}
+	return bm.rateSum[m] / float64(bm.rateCount[m])
+}
+
+func (bm *broadcasterMetrics) incErrorCount() map[string]int {
+	bm.mu.Lock()
+	bm.mu.Unlock()
+
+	return bm.incErrs
```
victorges, line 198: We should probably return a copy of this map, since after returning it from this function it would be unprotected by the lock.

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+
+	if bm.rateCount[m] <= 0 {
+		return 0
+	}
+	return bm.rateSum[m] / float64(bm.rateCount[m])
+}
+
+func (bm *broadcasterMetrics) incErrorCount() map[string]int {
+	bm.mu.Lock()
+	bm.mu.Unlock()
```
victorges, line 196: Missing the `defer`
```suggestion
	defer bm.mu.Unlock()
```

cmd/orch-tester/orch_tester.go
```diff
@@ -68,6 +70,134 @@ type broadcasterConfig struct {
...
+	bm.mu.Lock()
+	defer bm.mu.Unlock()
+
+	if bm.rateCount[m] <= 0 {
+		return 0
+	}
+	return bm.rateSum[m] / float64(bm.rateCount[m])
+}
+
+func (bm *broadcasterMetrics) incErrorCount() map[string]int {
```
victorges, line 194: This function name implied to me that it would be incrementing the error count, when it actually returns it. WDYT of just calling it `errorCounts` or something like that?

cmd/orch-tester/orch_tester.go
```diff
@@ -135,6 +265,13 @@ func main() {
 		bcastHost = *broadcaster
 	}
 
+	var embeddedBcastMetrics *broadcasterMetrics
+	if *streamTester == "" && *broadcaster == "" && *metrics == "" {
+		glog.Infof("Using embedded broadcaster metrics")
+		embeddedBcastMetrics = &broadcasterMetrics{}
```
victorges, line 271: Should probably use the constructor instead of a raw struct value. (or if we never use the constructor, maybe don't even need to have it)

cmd/orch-tester/orch_tester.go
```diff
@@ -470,75 +587,52 @@ func (s *streamerClient) getFinishedStats(mid string) (*streamerModel.Stats, err
...
-		}
-	}
-	return valFloat, nil
-}
 
-func (s *streamerClient) avgUploadTime() (float64, error) {
-	val, err := s.queryVectorMetric(
-		"rate(livepeer_upload_time_seconds_sum[1m])/rate(livepeer_upload_time_seconds_count[1m])",
-	)
+	query := fmt.Sprintf("sum(rate(livepeer_%s_sum[1m]))/sum(rate(livepeer_%s_count[1m]))", metric, metric)
```
victorges, line 595: Hmm previously it didn't even have the `sum` right? I think it makes sense to have it, but any idea how it worked before? Is it because we always had a single instance of the orch-tester broadcaster anyway? (how does it not mix the metrics with regular non-orch-tester metircs?)

victorges (approved):
LGTM! 

There are some smaller issues that may even be bugs, but they're not critical. So I can take another look if you'd like after the updates, just lmk if so. Most comments are nits tho.

cmd/orch-tester/orch_tester.go
leszko: Good idea. Done in 9e7bd6f5025836870000d1cae8994bb23c467a41

cmd/orch-tester/orch_tester.go
leszko: No, we don't reset all the fields, only the rate fields. The other fields (`lastSum`, `lastCount`, `lastErrs`) always keep the latest accumulated value from exported metrics. That is because census always aggregate metrics, so we need to know the last measurement. Metrics are tricky....

cmd/orch-tester/orch_tester.go
leszko: It could, but I'm not sure of the benefits of this. I think it's better to keep the scope limited to one function if they are not used anywhere else.

cmd/orch-tester/orch_tester.go
leszko: ⬆️

cmd/orch-tester/orch_tester.go
leszko: No, actually not. It's tricky, but census always gives the sum of all the previous metrics. So, in result, if we use these values as you mentioned, for the given O, we would take the values for all the past O measurements.

In other words, your code would correspond to the prometheus query like this one:
```
sum(livepeer_metricname_sum)/sum(livepeer_metricname_count)
```

While we actually want to scope it to one orchestrator only. So, a query like this (`1m` might not be exactly accurate, but that's what worked):

```
sum(rate(livepeer_metricname_sum[1m]))/sum(rate(livepeer_metricname_count[1m]))
```

cmd/orch-tester/orch_tester.go
leszko: ⬆️

cmd/orch-tester/orch_tester.go
leszko: You're right! Updated in d7abb3df43d95c636324b73ea956389f0a8c6622

cmd/orch-tester/orch_tester.go
leszko: Yes, makes sense. Updated in f3370009f3b806ace8229aa94ad321709948ddb0

cmd/orch-tester/orch_tester.go
leszko: Yeah, we don't need it. Removed.

cmd/orch-tester/orch_tester.go
leszko: Hmm, so the story is like this. We had 4 different metrics here:
- `transcode_overall_latency_seconds` => We already used the `sum`, so the query didn't change. The metric is exported with different tags, so we need to sum over different tags
- `upload_time_seconds` => We didn't use `sum`, but it was a bug [1]. The metric is exported with different tags, so we should sum it over different tags
- `source_segment_duration_seconds`, `download_time_seconds` => This metrics are not exported with tags, so technically we don't need to sum them; but `sum` does not harm and will work even if we introduce tags for these metrics at some point

[1] https://github.com/livepeer/stream-tester/issues/138

cmd/orch-tester/broadcaster_metrics.go
```diff
@@ -0,0 +1,126 @@
...
+	if len(rows) <= 0 {
+		return
+	}
+
+	bm.handleDistributionMetrics(rows, name)
+	bm.handleErrorCountMetrics(rows, name)
+}
+
+func (bm broadcasterMetrics) handleDistributionMetrics(rows []*view.Row, name string) {
+	supportedMetrics := map[string]bool{
```
red-0ne, line 46: Why isn't this a package variable?

red-0ne (approved):
Asked one side question but otherwise LGTM.

cmd/orch-tester/orch_tester.go
leszko: Ahh, no you're right, we need the constructor. Added it back and used it.

cmd/orch-tester/broadcaster_metrics.go
leszko: It could be, but I think it's better to limit the scope to the function if it's only used inside that function.

PR #174, opened 2022-07-11
workflow/docker: github matrix array to deal with docker images

hjpotter92:
Reduces multiple workflow actions to a single manageable manifest.


.github/workflows/docker.yaml
```diff
@@ -0,0 +1,77 @@
...
+        uses: livepeer/action-gh-release-tags@v0
+        with:
+          always-latest-on-branch: master
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          build-args: |
+            version=${{ steps.tags.outputs.version }}
```
victorges, line 71: Ooh, so we need the tags step only for the version build arg now?

victorges (commented):
Looks awesome! Just a remark on the tag filter since I think we might wanna keep it as "notify every time" if that doesn't create any problems.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        arrays: [
+          {name: 'loadtester', file:'Dockerfile.load-tester'},
+          {name: 'record-tester', file:'Dockerfile.record-tester'},
+          {name: 'orch-tester', file:'Dockerfile.orch-tester'},
+          {name: 'mist-api-connector', file:'Dockerfile.mist api-connector'},
+          {name: 'streamtester', file:'Dockerfile'},
```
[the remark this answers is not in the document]
hjpotter92: yep. it generates a set of 5 separate tests

.github/workflows/docker.yaml
```diff
@@ -73,5 +75,12 @@ jobs:
...
 
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: match-tag
+        with:
+          text: ${{ github.ref_name }}
+          regex: '^(master|main|v[0-9]+\.\d+\.\d+)$'
+
       - name: Notify that new build has been uploaded
+        if: ${{ steps.match-tag.outputs.match != '' }}
         run: curl -X POST https://holy-bread-207a.livepeer.workers.dev
```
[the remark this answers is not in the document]
hjpotter92: i wasn't sure how the worker notifications work, and assumed that the current docker tag might get deployed, which is not wanted in all cases.

> just have it be a no-op most of the time since the updated images don't matter to the currently deployed infra.

that makes it easier. i can remove the `match-tag` entirely.

.github/workflows/docker.yaml
hjpotter92: yeah! i couldn't find an alternative to this yet :(

hjpotter92: yeah! i couldn't find an alternative to this yet :(

victorges (approved):
LGTM!

PR #184, opened 2022-07-18
record-tester: Isolate stream health alerts from all others

victorges:
It is the most verbose of the tests and it is not 100% critical, so we shouldn't be paging
the on-call engineer with high urgency when they fire.

By keeping them together with all the other alerts, it also makes the other alerts less trusted
due to alert fatigue, and we end up ignoring real incidents.

On a higher level, we have moved away from the idea of the current Stream Health architecture
being the one and only source of truth for the Livepeer universe. So it is OK to start moving this
more to the background for now.


iameli (approved):
LGTM! `sortErrs()` is a cool function

PR #187, opened 2022-08-17
docker: Move all `Dockerfile`s in a contained directory

hjpotter92:
Revert `orch-tester` to use livepeer's build as base image since livepeer/go-livepeer#2512 is closed.


victorges (approved):
Nice!

PR #189, opened 2022-08-30
mapic: don't overwrite existing stream definitions

iameli:
So we can actually tweak things about those streams in production configuration without having to run through mapic configuration every dang time.


internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -1126,13 +1126,13 @@ func (mc *mac) SetupTriggers(ownURI string) error {
 		apiURL := mc.lapi.GetServer() + "/api/stream/" + mc.baseStreamName
 		presets := []string{"P144p30fps16x9"}
 		// base stream created with audio disabled
```
victorges, line 1128: Oooohhh could this be the root cause for those audio-less recordings we've seen a couple of times?

victorges (approved):
LGTM

PR #191, opened 2022-09-06
vod: upload tester

gioelecerati:


internal/app/vodtester/vodtester_app.go
```diff
@@ -144,11 +145,107 @@ func (vt *vodTester) Start(fileUrl string, taskPollDuration time.Duration) (int,
 		}
 	}
 
+	filePath := "/tmp/uploadtester.mp4"
+	resp, err := http.Get(fileUrl)
```
victorges, line 149: As we chatted on Discord, instead of downloading a new file we should just receive the `fileName` as args to the `Start` function, just like we already have for the `ContinuousRecordTester`: https://github.com/livepeer/stream-tester/blob/adbe3a0cf54c7eb79f61eee092b7080e26720508/cmd/recordtester/recordtester.go#L275

The `ContinuousVodTester` is right below that, and should also receive the `fileName` as arg and pass it down to this `VodTester` here. Then we only need to open the file instead of re-downloading with HTTP.

We also chatted this offline, but FTR: this file is already present in the Docker image as well, so we don't need the extra bandwidth to download it on every execution of the test.

internal/app/vodtester/vodtester_app.go
```diff
@@ -144,11 +145,107 @@ func (vt *vodTester) Start(fileUrl string, taskPollDuration time.Duration) (int,
...
+
+	if err != nil {
+		glog.Errorf("Error downloading fileUrl=%s err=%v", fileUrl, err)
+	}
+	defer resp.Body.Close()
+
+	out, err := os.Create(filePath)
+	if err != nil {
+		glog.Errorf("Error creating file filePath=%s err=%v", filePath, err)
+	}
```
victorges, line 159: Missing returns after handling the errors (tho this code should go away if you replace with the built-in file).

internal/app/vodtester/vodtester_app.go
```diff
@@ -144,11 +145,107 @@ func (vt *vodTester) Start(fileUrl string, taskPollDuration time.Duration) (int,
...
+		glog.Errorf("Error processing asset assetId=%s taskId=%s", uploadAsset.ID, uploadTask.ID)
+		return 254, err
+	}
+
 	glog.Info("Done VOD Test")
 
 	return 0, nil
 }
 
+func (vt *vodTester) checkAssetProcessing(taskPollDuration time.Duration, uploadAsset api.Asset, uploadTask api.Task) error {
```
victorges, line 224: Hmm, even better would be implementing a helper that works on top of a `Task` instead of an `Asset`, and then reuse it on the regular `import` (upload with URL) tests above as well. WDYT?

internal/app/vodtester/vodtester_app.go
```diff
@@ -144,11 +145,107 @@ func (vt *vodTester) Start(fileUrl string, taskPollDuration time.Duration) (int,
...
+	defer resp.Body.Close()
+
+	out, err := os.Create(filePath)
+	if err != nil {
+		glog.Errorf("Error creating file filePath=%s err=%v", filePath, err)
+	}
+	defer out.Close()
+
+	assetName = fmt.Sprintf("vod_test_asset_%s_%s", hostName, time.Now().Format("2006-01-02T15:04:05Z07:00"))
+	requestUpload, err := vt.lapi.RequestUpload(assetName)
```
victorges, line 163: It would be good to perform each of these tests in parallel as well, so we don't take the time of 3 uploads+processes for each continuous test iteration. WDYT?

I think the steps for that could be:
 - Create a helper for each of the tests being performed here (1 upload with url that we had before, 2 direct upload and 3 resumable upload)
 - Use a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) or even better an [ErrGroup](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) to perform them in parallel and return on errors from any of them.

You can ignore that `int` "error code" and just return the same for simplicity. I don't really think they are used for anything, just a tech debt we still have around here.

victorges (approved):
LGTM!

internal/app/vodtester/vodtester_app.go
gioelecerati: Sounds good! I moved everything into a `checkTaskProcessing` and removed this function

victorges (commented):
nice!

internal/app/vodtester/vodtester_app.go
```diff
@@ -44,56 +45,76 @@ func NewVodTester(gctx context.Context, opts VodTesterOptions) IVodTester {
...
 
-	importAsset, err := vt.importFromUrlTester(vodImportUrl, taskPollDuration, assetName)
+	eg.Go(func() error {
 
-	if err != nil {
-		glog.Errorf("Error importing asset from url=%s err=%v", vodImportUrl, err)
-		return 0, fmt.Errorf("error importing asset from url=%s: %w", vodImportUrl, err)
-	}
+		hostName, _ := os.Hostname()
+		assetName := fmt.Sprintf("vod_test_asset_%s_%s", hostName, time.Now().Format("2006-01-02T15:04:05Z07:00"))
```
victorges, line 53: WDYT of giving specific names for these, not only "test_asset"

This could be `vod_test_upload_url_...` instead (can ommit the `asset` its redundant)

Others would be `vod_test_upload_direct` and `vod_test_upload_resumable`

victorges (approved):
LGTM

PR #192, opened 2022-09-06
Add redirection to loadtester (and record-tester)

leszko:


iameli (approved):
Confirmed this seems to fix the problem with playback.livepeer.fun, nice! Thinking the only blocker left is the playlist download duplication stuff.

leszko:
> Confirmed this seems to fix the problem with playback.livepeer.fun, nice! Thinking the only blocker left is the playlist download duplication stuff.

Yes, created a separate GH Issue for that and I'll work on it.

https://github.com/livepeer/livepeer-infra/issues/943

ya7ya (approved):

PR #197, opened 2022-09-15
vod: vodloadtester (DirectUploads, ResumableUploads, Import)

gioelecerati:
- [X] Direct uploads load test
- [x] Resumable uploads load test
- [X] Import from url load test
- [X] Import from a .json file containing a list of remote files (ipfs, https)
- [X] Upload files from a folder
- [X] Save results with success / fail in steps, with error messages coming from the pipeline

Notion doc for usage:
https://www.notion.so/livepeer/How-to-run-a-VOD-load-test-using-vodloadtester-0b3db771393a46a384ce7e05b614150a#4fa99de44d34404fb134c515c48faf93
```
Usage of vodloadtester:
  -api-server string
    	Server of the Livepeer API to be used (default "origin.livepeer.monster")
  -api-token string
    	Token of the Livepeer API to be used
  -config string
    	config file (optional)
  -delay-between-groups duration
    	Delay between starting group of uploads (default 10s)
  -direct
    	Launch direct upload test
  -file string
    	File to upload or url to import. Can be either a video or a .json array of objects with a url key
  -folder string
    	Folder with files to upload. The tester will search in the folder for files with the following extensions: mp4
  -import
    	Launch import from url test
  -keep-assets
    	Keep assets after test
  -output-path string
    	Path to output result .ndjson file (default "/tmp/results.ndjson")
  -probe-data
    	Write object data in results when importing a JSON file
  -resumable
    	Launch tus upload test
  -sim uint
    	Number of simulteneous videos to upload or import (batch size) (default 1)
  -task-check
    	Check task processing
  -task-poll-time uint
    	Task check poll time in seconds (default 5)
  -task-timeout uint
    	Task check timeout in seconds (default 500)
  -v int
    	Log verbosity.  {4|5|6} (default 5)
  -version
    	Print out the version
  -video-amt uint
    	How many video to upload or import (default 1)
```

Example output:
```
{"start_time":"2022-09-16T20:30:29.171494319+02:00","end_time":"2022-09-16T20:30:50.362233778+02:00","runnerInfo":"Hostname gioele-XPS-15-9510 OS linux","kind":"resumable","assetId":"f5e5cbd3-8179-4a65-acb8-29707f96f780","taskId":"4268d966-3819-4ec2-81c6-4c6da5691e99","requestUploadSuccess":1,"uploadSuccess":1,"taskCheckSuccess":1}
{"start_time":"2022-09-16T20:30:29.171489196+02:00","end_time":"2022-09-16T20:30:50.393028209+02:00","runnerInfo":"Hostname gioele-XPS-15-9510 OS linux","kind":"resumable","assetId":"3a03fafc-a186-45c9-a6d6-c238c0f419a9","taskId":"a44ed917-40ec-4308-baef-353d7f1d6b07","requestUploadSuccess":1,"uploadSuccess":1,"taskCheckSuccess":1}
{"start_time":"2022-09-16T20:30:44.172579442+02:00","end_time":"2022-09-16T20:31:04.380322533+02:00","runnerInfo":"Hostname gioele-XPS-15-9510 OS linux","kind":"resumable","assetId":"76b2b2dc-9473-4512-80bf-a8359b3ecf6f","taskId":"5e5fd631-8e7f-4724-9590-e92fcd31d60e","requestUploadSuccess":1,"uploadSuccess":1,"taskCheckSuccess":1}
{"start_time":"2022-09-16T20:30:44.172512413+02:00","end_time":"2022-09-16T20:31:04.405383898+02:00","runnerInfo":"Hostname gioele-XPS-15-9510 OS linux","kind":"resumable","assetId":"b0c168c0-177a-49f6-87d3-09966d41e5d0","taskId":"a12b9672-3d7e-4bb1-a037-2c90d101d795","requestUploadSuccess":1,"uploadSuccess":1,"taskCheckSuccess":1}
{"start_time":"2022-09-16T20:30:59.175543883+02:00","end_time":"2022-09-16T20:31:14.582551084+02:00","runnerInfo":"Hostname gioele-XPS-15-9510 OS linux","kind":"resumable","assetId":"c5ea0148-8c2f-4760-a324-e432ba112bc0","taskId":"1571a2a9-17b5-4b6a-b95e-625f3e07dd68","requestUploadSuccess":1,"uploadSuccess":1,"taskCheckSuccess":1}
```

### DIFF
--- a/cmd/vodloadtester/vodloadtester.go
+++ b/cmd/vodloadtester/vodloadtester.go
@@ -95,7 +95,7 @@ func main() {
 	vFlag.Value.Set(fmt.Sprintf("%d", cliFlags.Verbosity))
 
 	hostName, _ := os.Hostname()
-	runnerInfo := fmt.Sprintf("Hostname %s OS %s IPs %v\n", hostName, runtime.GOOS, utils.GetIPs())
+	runnerInfo := fmt.Sprintf("Hostname %s OS %s", hostName, runtime.GOOS)
 	fmt.Printf("Compiler version: %s %s\n", runtime.Compiler, runtime.Version())
 	fmt.Print(runnerInfo)
 
--- a/cmd/vodloadtester/vodloadtester.go
+++ b/cmd/vodloadtester/vodloadtester.go
@@ -134,10 +134,11 @@ func main() {
 
 	if cliFlags.Import {
 		if cliFlags.DirectUpload || cliFlags.ResumableUpload {
-			glog.Fatal("Cannot use -import with either -direct or -resumable")
+			glog.Infof("Cannot use -import with either -direct or -resumable, doing only the import task")
 		}
 		fileName = cliFlags.Filename
 		vt.importFromUrlTest(fileName, runnerInfo)
+		return
 	}
 
 	if cliFlags.DirectUpload || cliFlags.ResumableUpload {
@@ -185,6 +186,7 @@ func (vt *vodLoadTester) directUploadLoadTest(fileName string, runnerInfo string
 				rndAssetName := fmt.Sprintf("load_test_direct_%s", randName())
 				requestedUpload, err := vt.requestUploadUrls(rndAssetName)
 				defer vt.lapi.DeleteAsset(requestedUpload.Asset.ID)
+				defer wg.Done()
 
 				if err != nil {
 					glog.Errorf("Error requesting upload urls: %v", err)
@@ -198,7 +200,7 @@ func (vt *vodLoadTester) directUploadLoadTest(fileName string, runnerInfo string
 					uploadTest.TaskID = requestedUpload.Task.ID
 				}
 
-				err = vt.uploadAsset(fileName, requestedUpload.Url)
+				err = vt.uploadAsset(fileName, requestedUpload.Url, false)
 
 				if err != nil {
 					glog.Errorf("Error uploading asset: %v", err)
@@ -217,7 +219,6 @@ func (vt *vodLoadTester) directUploadLoadTest(fileName string, runnerInfo string
 
 				}
 				vt.writeResultNdjson(uploadTest)
-				wg.Done()
 			}()
 		}
 		time.Sleep(vt.cliFlags.StartDelayDuration)
@@ -242,6 +243,7 @@ func (vt *vodLoadTester) resumableUploadLoadTest(fileName, runnerInfo string) {
 				rndAssetName := fmt.Sprintf("load_test_resumable_%s", randName())
 				requestedUpload, err := vt.requestUploadUrls(rndAssetName)
 				defer vt.lapi.DeleteAsset(requestedUpload.Asset.ID)
+				defer wg.Done()
 
 				if err != nil {
 					glog.Errorf("Error requesting upload urls: %v", err)
@@ -257,7 +259,7 @@ func (vt *vodLoadTester) resumableUploadLoadTest(fileName, runnerInfo string) {
 
 				uploadUrl := requestedUpload.TusEndpoint
 
-				err = vt.uploadAssetResumable(uploadUrl, fileName)
+				err = vt.uploadAsset(fileName, uploadUrl, true)
 
 				if err != nil {
 					glog.Errorf("Error on resumable upload: %v", err)
@@ -276,7 +278,6 @@ func (vt *vodLoadTester) resumableUploadLoadTest(fileName, runnerInfo string) {
 
 				}
 				vt.writeResultNdjson(uploadTest)
-				wg.Done()
 			}()
 		}
 		time.Sleep(vt.cliFlags.StartDelayDuration)
@@ -303,6 +304,7 @@ func (vt *vodLoadTester) importFromUrlTest(url string, runnerInfo string) {
 				rndAssetName := fmt.Sprintf("load_test_import_%s", randName())
 				asset, task, err := vt.lapi.ImportAsset(url, rndAssetName)
 				defer vt.lapi.DeleteAsset(asset.ID)
+				defer wg.Done()
 
 				if err != nil {
 					glog.Errorf("Error importing asset: %v", err)
@@ -326,7 +328,6 @@ func (vt *vodLoadTester) importFromUrlTest(url string, runnerInfo string) {
 					uploadTest.EndTime = time.Now()
 				}
 				vt.writeResultNdjson(uploadTest)
-				wg.Done()
 			}()
 		}
 		time.Sleep(vt.cliFlags.StartDelayDuration)
@@ -350,7 +351,7 @@ func (vt *vodLoadTester) writeResultNdjson(uploadTest uploadTest) {
 	}
 }
 
-func (vt *vodLoadTester) uploadAsset(fileName string, uploadUrl string) error {
+func (vt *vodLoadTester) uploadAsset(fileName string, uploadUrl string, resumable bool) error {
 
 	file, err := os.Open(fileName)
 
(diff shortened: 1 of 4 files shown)


gioelecerati:
Fixes https://github.com/livepeer/studio/issues/1211

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,423 @@
...
+	APIToken    string
+	Filename    string
+	VideoAmount uint
+	OutputPath  string
+
+	TestDuration       time.Duration
+	StartDelayDuration time.Duration
+}
+
+type vodLoadTester struct {
```
victorges, line 41: Could we move this struct to its own package under `internal/app/vodloadtester`? That's the pattern we've been following in this project, and I think it's indeed cleaner to separate the business logic of the load tester from the code that only deals with the "command line interface" parsing/serializing of results etc.

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,423 @@
...
+
+	if cliFlags.Import {
+		if cliFlags.DirectUpload || cliFlags.ResumableUpload {
+			glog.Fatal("Cannot use -import with either -direct or -resumable")
+		}
+		fileName = cliFlags.Filename
+		vt.importFromUrlTest(fileName, runnerInfo)
+	}
+
+	if cliFlags.DirectUpload || cliFlags.ResumableUpload {
```
victorges, line 143: Should this be an `else if`, or should we at least `return` from the above branch? From the other checks it seems like it's either one or ther other here.
```suggestion
		vt.importFromUrlTest(fileName, runnerInfo)
		return
	}

	if cliFlags.DirectUpload || cliFlags.ResumableUpload {
```
or
```suggestion
		vt.importFromUrlTest(fileName, runnerInfo)
	} else if cliFlags.DirectUpload || cliFlags.ResumableUpload {
```

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func (vt *vodLoadTester) resumableUploadLoadTest(fileName, runnerInfo string) {
+	wg := &sync.WaitGroup{}
+
+	for i := 0; i < int(vt.cliFlags.VideoAmount); i += int(vt.cliFlags.Simultaneous) {
+		for j := 0; j < int(vt.cliFlags.Simultaneous); j++ {
+			fmt.Printf("rUpload video %d/%d\n", i+j+1, vt.cliFlags.VideoAmount)
+			wg.Add(1)
+			go func() {
```
victorges, line 236: I think that with this logic you are actually doing `VideoAmount * Simultaneous` concurrent uploads. What you want is probably to spawn `Simultaneous` routines and iterate (together) over `VideoAmount` videos, which you can do using a channel. We can pair on this if you want, but this might also help: https://go.dev/blog/pipelines

victorges (commented):
sorry had to send a partial review tuesday on a rush. coming back to this now

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,423 @@
...
+							uploadTest.TaskCheckSuccess = 0
+							uploadTest.ErrorMessage = err.Error()
+						} else {
+							uploadTest.TaskCheckSuccess = 1
+						}
+					}
+
+				}
+				vt.writeResultNdjson(uploadTest)
+				wg.Done()
```
victorges, line 279: You **must** do `defer wg.Done()` in the beginning of this function instead of calling it manually in the end, like all the `WaitGroup` code examples out there. No real reason not to use `defer` on this case, since it's much simpler and you won't need to handle all the "alternative exits" of your function (you are already missing one now).

On the current version, you already have a source of a deadlock on line 251, in which you end the function without ever calling `Done()`, meaning the test would never end. This is exactly the reason to use `defer` instead, which will avoid such kinds of bugs. Sorry `go` won't automatically tattle on those on compilation like Rust does 😆

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,423 @@
...
+
+}
+
+func (vt *vodLoadTester) importFromUrlTest(url string, runnerInfo string) {
+	wg := &sync.WaitGroup{}
+
+	for i := 0; i < int(vt.cliFlags.VideoAmount); i += int(vt.cliFlags.Simultaneous) {
+		for j := 0; j < int(vt.cliFlags.Simultaneous); j++ {
+			fmt.Printf("Importing video %d/%d\n", i+j+1, vt.cliFlags.VideoAmount)
+			wg.Add(1)
```
victorges, line 295: This has the same problems stated on the above function. Gotta fix the concurrency and the usage of `sync.WaitGroup`.

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	err = vt.lapi.ResumableUpload(url, file)
+
+	if err != nil {
+		fmt.Printf("error on resumable upload asset: %v", err)
+	}
+
+	return err
+}
```
victorges, line 387: **nit:** These are so similar, could even be a single func with an opt, WDYT?
```suggestion
func (vt *vodLoadTester) uploadAsset(fileName string, uploadUrl string, resumable bool) error {
	file, err := os.Open(fileName)
	if err != nil {
		fmt.Printf("error opening file %s: %v\n", fileName, err)
		return err
	}

        if resumable {
          err = vt.lapi.ResumableUpload(uploadUrl, file)
        } else {
	  err = vt.lapi.UploadAsset(vt.ctx, uploadUrl, file)
	}
	if err != nil {
		fmt.Printf("error uploading asset: %v\n", err)
		return err
	}

	return nil
}
```

victorges (commented):
Overall LGTM, but the concurrency logic needs some fixes

cmd/vodloadtester/vodloadtester.go
gioelecerati: Yeah, I was doing the check of not mixing the two possible inputs, but it's file like this, I added a return in the import

cmd/vodloadtester/vodloadtester.go
gioelecerati: Thank you, it makes much sense! <3 `defer` is actually very cool

cmd/vodloadtester/vodloadtester.go
gioelecerati: Good point!

cmd/vodloadtester/vodloadtester.go
gioelecerati: Sounds good, although I think something is missing here. With the previous version, I could specify a `StartDelayDuration` between groups, so that I could fire groups of video to be imported separated by a specified delay. 

This way instead, subsequent videos are uploaded only on task completion, am I wrong?

gioelecerati: For example, I cannot make batches of 5 videos to be uploaded every 30 seconds

victorges (approved):
LGTM!

cmd/vodloadtester/vodloadtester.go
```diff
@@ -0,0 +1,523 @@
...
+			}()
+		}
+		time.Sleep(vt.cliFlags.StartDelayDuration)
+	}
+
+	wg.Wait()
+
+}
+
+func (vt *vodLoadTester) importFromJSON(jsonFile string, runnerInfo string) {
```
victorges, line 317: Coooool

PR #220, opened 2022-11-10
record-tester: added access control testing

### LINKED ISSUE #204: Create continuous test for access controlled media (secure playback)

gioelecerati:
Fixes #204

### DIFF
--- a/internal/app/recordtester/recordtester_app.go
+++ b/internal/app/recordtester/recordtester_app.go
@@ -178,17 +178,13 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 	glog.V(model.SHORT).Infof("RTMP: %s", rtmpURL)
 	glog.V(model.SHORT).Infof("MEDIA: %s", mediaURL)
 
-	if rt.accessControl {
-		if rt.signingKey == "" || rt.publicKey == "" {
-			return 2, errors.New("test access control is enabled but no signing key or public key is provided")
+	if rt.accessControl && (rt.signingKey != "" && rt.publicKey != "") {
+		token, err := rt.signJwt(stream)
+		if err != nil {
+			return 1, err
 		}
-
-		es, err := rt.testAccessControl(stream, mediaURL, fileName, rtmpURL, testDuration)
-		if err == nil {
-			rt.lapi.DeleteStream(stream.ID)
-		}
-
-		return es, err
+		mediaURL = fmt.Sprintf("%s?jwt=%s", mediaURL, token)
+		glog.V(model.VERBOSE).Infof("URL with access control for stream id=%s playbackId=%s name=%s mediaURL=", stream.ID, stream.PlaybackID, streamName, mediaURL)
 	}
 
 	if rt.useHTTP {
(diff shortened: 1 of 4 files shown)


internal/app/recordtester/recordtester_app.go
```diff
@@ -156,6 +184,19 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 	mediaURL := fmt.Sprintf("%s/%s/index.m3u8", ingest.Playback, stream.PlaybackID)
 	glog.V(model.SHORT).Infof("RTMP: %s", rtmpURL)
 	glog.V(model.SHORT).Infof("MEDIA: %s", mediaURL)
+
+	if rt.accessControl && (rt.signingKey != "" && rt.publicKey != "") {
```
victorges, line 188: Having `accessControl` and not the signing key should be an explicit fatal error. We should not ignore the setting because another required option was not provided, it is better to just fail the whole test.

```suggestion
	if rt.accessControl {
	  if rt.signingKey == "" || rt.publicKey == "" {
	    msg := "Signing key and public key are required for access control test"
	    glog.Fatal(msg)
	    return 1, errors.New(msg) // (Fatal log above already exits the process but you know just in case, leave it clear here that we end with an error)
	  }
```

internal/app/recordtester/recordtester_app.go
```diff
@@ -156,6 +184,19 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+		token, err := rt.signJwt(stream)
+		if err != nil {
+			return 1, err
+		}
+		mediaURL = fmt.Sprintf("%s?jwt=%s", mediaURL, token)
+		glog.V(model.VERBOSE).Infof("URL with access control for stream id=%s playbackId=%s name=%s mediaURL=%s", stream.ID, stream.PlaybackID, streamName, mediaURL)
+	} else {
+		glog.Warningf("No access control for stream id=%s playbackId=%s name=%s mediaURL=%s", stream.ID, stream.PlaybackID, streamName, mediaURL)
+		return 2, nil
+	}
```
victorges, line 198: No reason to log this as a warning or returning. Executing a test without access control is perfectly acceptable. And we already log above when we use access control so no need to log when we don't.
```suggestion
	}
```

internal/app/recordtester/recordtester_app.go
```diff
@@ -280,51 +321,41 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
-	if es == 0 {
-		rt.lapi.DeleteStream(stream.ID)
-		// exit(0, fileName, *fileArg, err)
-	}
+	glog.Info("Done Record Test")
 
-	// uploader := testers.NewRtmpStreamer(gctx, rtmpURL)
-	// uploader.StartUpload(fileName, rtmpURL, -1, 30*time.Second)
-	return es, err
+	return 0, err
```
victorges, line 358: **nit:** We know `err` is `nil` here so we can be explicit. Just so one doesn't start wondering here where this `err` could be instead of `nil` (from what I checked, it can't)
```suggestion
	return 0, nil
```

internal/app/recordtester/recordtester_app.go
```diff
@@ -389,6 +420,31 @@ func (rt *recordTester) isCancelled() error {
...
+
+	if err != nil {
+		glog.Errorf("Unable to parse provided signing key for access control signingKey=%s", rt.signingKey)
+	}
+
+	token, err := unsignedToken.SignedString(pk)
+
+	if err != nil {
+		glog.Errorf("Unable to sign JWT with provided private key for access control signingKey=%s", rt.signingKey)
+	}
```
victorges, line 443: We need to return all these errors. Cannot just log.
```suggestion
	if err != nil {
		glog.Errorf("Unable to parse provided signing key for access control signingKey=%s", rt.signingKey)
		return "", err
	}

	token, err := unsignedToken.SignedString(pk)

	if err != nil {
		glog.Errorf("Unable to sign JWT with provided private key for access control signingKey=%s", rt.signingKey)
		return "", err
	}
```

victorges (approved):
LGTM

PR #221, opened 2022-11-13
feat: Make mapic stateless

victorges:
This is to remove the assumption on mapic code that it runs forever alongside mist and
allow for it to be restarted.

For this, removed 2 main aspects of the code where mapic behaved in a stateful way:
 - etcd routes: we don't even use these anymore, so just deleted all the code about it
 - stream info: made this inot some kind of cache instead, fetching the stream from the API if it's not in the memory map yet


cmd/mist-api-connector/mist-api-connector.go
```diff
@@ -44,18 +44,20 @@ func main() {
...
-	etcdKey := fs.String("etcd-key", "", "ETCD client certificate key file name")
 	amqpUrl := fs.String("amqp-url", "", "RabbitMQ url")
 	ownRegion := fs.String("own-region", "", "Identifier of the region where the service is running, used for mapping external data back to current region")
 	_ = fs.String("config", "", "config file (optional)")
+	// Below are some deprecated flags.
+	// Keep them around for backward compatibility on deploys.
+	_ = fs.String("etcd-endpoints", "", "DEPRECATED")
+	_ = fs.String("etcd-cacert", "", "DEPRECATED")
+	_ = fs.String("etcd-cert", "", "DEPRECATED")
+	_ = fs.String("etcd-key", "", "DEPRECATED")
```
iameli, line 60: I'm cool with this but we don't need them on deploys so I kind of doubt that anyone else does either

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -821,185 +660,45 @@ func (mc *mac) handleDefaultStreamTrigger(w http.ResponseWriter, r *http.Request
...
+	}()
 }
 
-func (mc *mac) recoverEtcdSessionOnce() error {
-	sess, err := newEtcdSession(mc.etcdClient)
-	if err != nil {
-		return err
-	}
-
+// must be called inside mu.Lock
```
iameli, line 674: Both functions shouldn't have this comment, correct?

internal/app/mistapiconnector/mistapiconnector_app.go
```diff
@@ -1224,70 +873,98 @@ func (mc *mac) startSignalHandler() {
...
+func (mc *mac) getPushUrl(stream *api.Stream, targetRef *api.MultistreamTargetRef) (*api.MultistreamTarget, string, error) {
+	target, err := mc.lapi.GetMultistreamTarget(targetRef.ID)
+	if err != nil {
+		return nil, "", fmt.Errorf("error fetching multistream target %s: %w", targetRef.ID, err)
+	}
+	// Find the actual parameters of the profile we're using
+	var videoSelector string
+	// Not actually the source. But the highest quality.
+	if targetRef.Profile == "source" {
+		videoSelector = "maxbps"
```
iameli, line 885: We just hit a bug with this in MistProcLivepeer where sometimes a transcoded rendition can be higher bitrate. @Thulinma is there a correct way for us to select just the source hre?

internal/app/mistapiconnector/mistapiconnector_app.go
victorges: Right! Moved to only the one below.

internal/app/mistapiconnector/mistapiconnector_app.go
victorges: FTR as we chatted about this, this is not a behavior change in this PR, it was only copied from the previous code. We can merge this as is and follow up with a fix for this later.

Thulinma: MistProcLivepeer, since an update we rolled out during the holidays, can no longer select its own renditions as source for itself. That _should_ handle this situation without any changes needed.

To answer the actual question: yes, sort-of. You can simply _not_ specify any specific track selector (leave it default value), and set the default track selector to `id_lth` to prioritize the lowest-ID track of the type, which for live streams is guaranteed to be the source (or one of the source tracks, at least, if there is more than one video track in the source already).

victorges (commented):
self-LGTM before merge

PR #234, opened 2022-12-28
recordtester: Use random serf memeber for playback if enabled.

### LINKED ISSUE #227: record-tester: learn to download a list of nodes from Serf
New flag on recordtester similar to `--ingest`, called `--serf-rpc`. If defined, we should connect to that endpoint to retrieve a list of nodes. We should then pick a node close to us.

hjpotter92:
Fixes #227 

---

Adds 3 new cli flags:

- `--serf-rpc-addr`: Used to connect to serf cluster and fetch list of current members
- `--use-serf`: Whether to really use serf cluster member for playback URL. Default is false.
- `--random-serf-member`:  Uses a random selection from serf members list for playback. Otherwise uses the first member from the list.


iameli (approved):
This is a good (and mergeable) first step, but I think we'll want two more complex things here:

1. "Pull count" parameter so you can pull from multiple nodes; each one will need to trigger a separate playback
2. We'll want this to be able to geolocate so we're not unnecessarily testing pulls across the world. I guess this could be implemented with Serf tags to add the latitude and longitude parameters, and then a `-max-distance` parameter that limits how far the randomized Serf nodes are checked. (I guess we'll also need local `-latitude` and `-longitude` parameters to make that happen)

hjpotter92:
> 1. "Pull count" parameter so you can pull from multiple nodes; each one will need to trigger a separate playback

added a `TODO` with unimplemented feature for this.

> 2. We'll want this to be able to geolocate so we're not unnecessarily testing pulls across the world. I guess this could be implemented with Serf tags to add the latitude and longitude parameters, and then a `-max-distance` parameter that limits how far the randomized Serf nodes are checked. (I guess we'll also need local `-latitude` and `-longitude` parameters to make that happen)

sounds like a plan. let me see if i can update serf tags.

----

merging this PR at the current stage. will implement geolocation and pull-count once serf tags are updated.

PR #238, opened 2023-01-05
Add Transcode API

leszko:
Adds testing for the Studio API `/api/transcode`.

Two scenarios are tested:
- Transcoding a file from a publicly available URL
- Transcoding a file from a private Storj bucket

Note that I added this the same way as VOD, so it's a continuous test, but I think we should consider running it more like acceptance test + smoke test. I don't think we need to run it because of the MediaConvert cost.

---------------------

**To productize it we'll need to**
- Create a Livepeer Testing Storj bucket
- Add it as a job (continuously running pod) to our k8 cluster
- (optionally) Consider cleaning up the bucket from time to time

-------------

**Notes to reviewers**
- All functional changes are included in one commit e88766bf84655cfc8729461fc14fc08b8bed2ffb
- Two other commits are refactoring, they improve code reusability, but decrease the encapsulation a little bit
- This related PR https://github.com/livepeer/go-api-client/pull/23 needs to be merged first and `go.mod` dep updated

### DIFF
--- a/internal/app/common/continuous_tester.go
+++ b/internal/app/common/continuous_tester.go
@@ -15,7 +15,7 @@ import (
 type (
 	IContinuousTester interface {
 		// Start test. Blocks until error.
-		Start(start func(ctx context.Context) (int, error), testDuration, pauseBetweenTests time.Duration) error
+		Start(start func(ctx context.Context) error, testDuration, pauseBetweenTests time.Duration) error
 		Cancel()
 		Done() <-chan struct{}
 	}
@@ -54,14 +54,14 @@ func NewContinuousTester(gctx context.Context, opts ContinuousTesterOptions, tes
 	return ct
 }
 
-func (ct *continuousTester) Start(start func(ctx context.Context) (int, error), testDuration, pauseBetweenTests time.Duration) error {
+func (ct *continuousTester) Start(start func(ctx context.Context) error, testDuration, pauseBetweenTests time.Duration) error {
 	messenger.SendMessage(fmt.Sprintf("Starting continuous %s test of %s", ct.name, ct.host))
 	for {
 		msg := fmt.Sprintf(":arrow_right: Starting %s %s test to %s", testDuration, ct.name, ct.host)
 		messenger.SendMessage(msg)
 
 		ctx, cancel := context.WithTimeout(ct.ctx, testDuration)
-		es, err := start(ctx)
+		err := start(ctx)
 		ctxErr := ctx.Err()
 		cancel()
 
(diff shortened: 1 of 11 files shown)


internal/app/common/continuous_tester.go
```diff
@@ -0,0 +1,144 @@
...
+			messenger.SendMessage(fmt.Sprintf("Continuous test of %s on %s cancelled", ct.name, ct.host))
+			return ct.ctx.Err()
+		} else if ctxErr != nil {
+			msg := fmt.Sprintf("Test of %s on %s timed out, potential deadlock! ctxErr=%q err=%q", ct.name, ct.host, ctxErr, err)
+			messenger.SendFatalMessage(msg)
+		} else if err != nil || es != 0 {
+			msg := fmt.Sprintf(":rotating_light: Test of %s on %s ended with err=%v errCode=%v", ct.name, ct.host, err, es)
+			messenger.SendFatalMessage(msg)
+			glog.Warning(msg)
+			ct.sendPagerdutyEvent(err)
```
iameli, line 78: I know you're probably following existing patterns in stream-tester, but I've been really trying to drive everything toward Grafana. Centralizing our alerting there  allows everything to be managed in one place, allows alerts to be tuned and silenced when necessary — very useful.

internal/app/common/continuous_tester.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/PagerDuty/go-pagerduty"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/stream-tester/messenger"
+)
+
+type (
+	IContinuousTester interface {
+		// Start test. Blocks until error.
+		Start(start func(ctx context.Context) (int, error), testDuration, pauseBetweenTests time.Duration) error
```
thomshutt, line 18: What's the thinking behind having an error code and an error object returned here?

internal/app/common/continuous_tester.go
leszko: I think centralizing everything with Grafana is a good idea. Though, I think we should change it in a separate PR. What do you think?

This PR is just to add the Transcode API support in the exactly the same way we support VOD API.

Btw. do you have any example of using Grafana in streamtester?

internal/app/common/continuous_tester.go
leszko: To be honest I don't know, I think it was copy paster from recordtester into the vod tester. I just refactored something we already had in VOD tester 🙃 

Saying that, when I looked again in the code, I think we return always `0` or `1` for the error code, so I think it's safe to remove. Then, removed.

thomshutt (approved):

internal/app/common/continuous_tester.go
leszko: Ok, merging this PR. We can do the refactoring to use Grafana (for both VOD and Storj) in a separate PR.

PR #251, opened 2023-01-24
Log Catalyst pipeline strategy

thomshutt:


internal/app/vodtester/vodtester_app.go
```diff
@@ -129,7 +129,7 @@ func (vt *vodTester) uploadViaUrlTester(vodImportUrl string, taskPollDuration ti
 		glog.Errorf("Error importing asset err=%v", err)
 		return nil, fmt.Errorf("error importing asset: %w", err)
 	}
-	glog.Infof("Importing asset taskId=%s outputAssetId=%s", importTask.ID, importAsset.ID)
+	glog.Infof("Importing asset taskId=%s outputAssetId=%s pipelineStrategy=%s", importTask.ID, importAsset.ID, vt.CatalystPipelineStrategy)
```
victorges, line 132: There are some other places where it is used in this file, specifically for direct upload tests (string search should be enough to find them). Would you mind adding the field to the corresponding logs as well?

victorges (approved):
LGTM with a suggestion, no need for rereview

internal/app/vodtester/vodtester_app.go
thomshutt: Sure!

PR #267, opened 2023-02-13
Add Transcode API with web3.storage into record-tester

leszko:


cmd/recordtester/recordtester.go
```diff
@@ -419,7 +420,7 @@ func main() {
 					TesterOptions:           ttOpts,
 				}
 				ctt := transcodetester.NewContinuousTranscodeTester(egCtx, cttOpts)
-				return ctt.Start(*fileArg, *transcodeBucketUrl, *testDuration, *taskPollDuration, *continuousTest)
+				return ctt.Start(*fileArg, *transcodeBucketUrl, *transcodeW3sProof, *testDuration, *taskPollDuration, *continuousTest)
```
victorges, line 423: I'd say the ideal place for most new config should be in the `cttOpts` instead (likely in the underlying `ttOpts`).

Honestly I don't really know why some of these fields were added in the `Start` function instead of options of the tester, but adding a field to the struct is much easier to do and maintain anyway.

internal/app/transcodetester/transcodetester_app.go
```diff
@@ -199,25 +240,66 @@ func (tt *transcodeTester) copyFileIntoInputBucket(inUrl string, os objectStore,
...
-func (tt *transcodeTester) checkRenditionFiles(os objectStore, path string) error {
-	svc := s3.New(newAwsSession(os))
-	_, err := svc.GetObjectWithContext(tt.Ctx, &s3.GetObjectInput{
-		Bucket: aws.String(os.bucket),
-		Key:    aws.String(path2.Join(path, "index.m3u8")),
-	})
-	return err
+func (tt *transcodeTester) checkRenditionFiles(task *api.Task, path string, os *objectStore) error {
+	pathsToCheck := []string{path2.Join(path, "index.m3u8")}
+	pathsToCheck = append(pathsToCheck, task.Output.TranscodeFile.Hls.Path)
```
victorges, line 258: cross-pr-conversation: oh i get it now

internal/app/transcodetester/transcodetester_app.go
```diff
@@ -199,25 +240,66 @@ func (tt *transcodeTester) copyFileIntoInputBucket(inUrl string, os objectStore,
...
+			_, err := svc.GetObjectWithContext(tt.Ctx, &s3.GetObjectInput{
+				Bucket: aws.String(os.bucket),
+				Key:    aws.String(path2.Join(p)),
+			})
+			return err
+		}
+	}
+
+	baseUrlStr := task.Output.TranscodeFile.BaseUrl
+	if baseUrlStr != "" {
```
victorges, line 276: nitnitnit:
```suggestion
	if baseUrlStr := task.Output.TranscodeFile.BaseUrl; baseUrlStr != "" {
```

internal/app/transcodetester/transcodetester_app.go
```diff
@@ -199,25 +240,66 @@ func (tt *transcodeTester) copyFileIntoInputBucket(inUrl string, os objectStore,
...
+		// if baseUrl is returned, then a file should be accessible by this URL
+		baseUrl, err := url.Parse(baseUrlStr)
+		if err != nil {
+			return err
+		}
+		if baseUrl.Scheme == "ipfs" {
```
victorges, line 282: I think we should probably err if the scheme is not IPFS tho. We can't expect a different scheme from a web3.storage export being tested here

internal/app/transcodetester/transcodetester_app.go
```diff
@@ -199,25 +240,66 @@ func (tt *transcodeTester) copyFileIntoInputBucket(inUrl string, os objectStore,
...
+			// for now, we only support IPFS via the web3.storage gateway
```
victorges, line 283: It's only going to be slower if we use another gateway, isn't it? They should all work though as the files are in the IPFS network.

internal/app/transcodetester/transcodetester_app.go
```diff
@@ -199,25 +240,66 @@ func (tt *transcodeTester) copyFileIntoInputBucket(inUrl string, os objectStore,
...
+			w3sLinkBaseUrl, err := url.Parse(fmt.Sprintf("https://%s.ipfs.w3s.link/", baseUrl.Host))
+			if err != nil {
+				return err
+			}
+			for _, p := range pathsToCheck {
+				u := w3sLinkBaseUrl.JoinPath(p)
+				resp, err := http.Get(u.String())
```
victorges, line 290: Does w3s.link support `HEAD`? If so, that might be preferred

victorges (approved):
LGTM

thomshutt (approved):
LGTM once the build is passing

internal/app/transcodetester/transcodetester_app.go
leszko: Right, updated.

internal/app/transcodetester/transcodetester_app.go
leszko: Good idea, updated.

internal/app/transcodetester/transcodetester_app.go
leszko: Not exactly. In general, they should always work, but it may take "up to 48h".

cmd/recordtester/recordtester.go
leszko: I think the reason is that `cttOpts` (and `ttOpts`) are shared options between different testers (vod and transcode) and everything we put inside `Start()` is specific to the given test.

Saying that, I agree it's better to have this in some kind of structs. I'll refactor it accordingly.

leszko: ok, I started the refactor, but then I realized that to refactor it properly it's rather a day or two, rather than an hour or two. So, I created the related GH Issue.

I'm not sure we want to spend time on that right now.

Here's the GH Issue: https://github.com/livepeer/stream-tester/issues/268
Here's the started refactor: https://github.com/livepeer/stream-tester/tree/rafal/refactor-recordtester

internal/app/transcodetester/transcodetester_app.go
victorges: Hmm, isn't that the time to get the filecoin deals instead?

PR #295, opened 2023-05-23
Check m3u8 and mp4 to contain only 1 video stream

leszko:


internal/utils/video.go
```diff
@@ -72,22 +67,23 @@ func GetVideoStartTimeDurFrames(segment []byte) (time.Duration, time.Duration, i
...
-		// glog.Infof("=======--- streams: %+v", strms)
 		for i, s := range strms {
 			if s == nil {
 				continue
 			}
 			if s.Type().IsVideo() {
-				videoIdx = int8(i)
-				break
+				if videoIdx == 0 {
+					videoIdx = int8(i)
```
victorges, line 76: Perhaps init `videoIdx` with `-1` so we can differentiate no video from video on first position? Or is that already handled in some way?

victorges (approved):

internal/utils/video.go
leszko: Yeah, good point, updated.

PR #298, opened 2023-06-06
Add stream and playback IDs to logging and errors

mjh1:
To make debugging easier


victorges (approved):
Great!

internal/app/recordtester/recordtester_app.go
```diff
@@ -142,11 +141,17 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 		}
 		break
 	}
+	// Make sure we include the Stream and Playback IDs on any error message
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("streamId=%s playbackId=%s %w", stream.ID, stream.PlaybackID, err)
```
victorges, line 147: I think either one or the other should be cleaner, since both are unique values. Otherwise the message on Discord/Pagerduty incident titel could get huge and get cropped. So WDYT of keeping only `playbackId` which is shorter as well?

mjh1: Yeah that sounds good 👍 will make it playbackID only

PR #328, opened 2023-11-09
Implement test by URL for webrtc load test

mjh1:


cmd/webrtc-load-tester/roles/orchestrator.go
```diff
@@ -269,6 +273,17 @@ func playerJobSpec(args loadTestArguments, region string, viewers int, playbackI
 	numTasks := viewers / args.Playback.ViewersPerWorker
 	timeout := args.TestDuration + 10*time.Minute
 
+	playbackURL := ""
+	if args.Playback.ManifestURL != "" {
+		u, err := url.Parse(args.Playback.ManifestURL)
+		if err != nil {
+			log.Println("failed to parse URL ", args.Playback.ManifestURL, err)
+		} else {
+			u = u.JoinPath("video+" + playbackID)
```
victorges, line 282: If it seems helpful, a more flexible alternative would be for the input `ManifestURL` to be an `fmt`-format based string where the playback ID gets injected. e.g. this case would be a configuration like `livepeer.com/hls/video+%s` which we build with `fmt.Sprintf()` here.

cmd/webrtc-load-tester/roles/orchestrator.go
```diff
@@ -277,8 +292,7 @@ func playerJobSpec(args loadTestArguments, region string, viewers int, playbackI
 		Args: []string{
 			"-base-url", args.Playback.BaseURL,
 			"-playback-id", playbackID,
-			// TODO: Support region/node-specific playback by building the playback URL here
-			// "-playback-url", stream.PlaybackURL,
```
victorges, line 281: Does the player role already support this flag?

victorges (approved):
LGTM

cmd/webrtc-load-tester/roles/orchestrator.go
mjh1: Yep seems like you implemented that already!

cmd/webrtc-load-tester/roles/orchestrator.go
mjh1: Yeah that makes sense actually and makes the code simpler

cmd/webrtc-load-tester/roles/orchestrator.go
```diff
@@ -138,7 +140,7 @@ func runLoadTest(ctx context.Context, args loadTestArguments) error {
 	}
 
 	glog.Infof("Stream created: %s", stream.ID)
-	glog.Infof("Access the stream at: %s", path.Join(args.APIServer, "/dashboard/streams", stream.ID))
+	glog.Infof("Access the stream at: https://%s", path.Join(args.APIServer, "/dashboard/streams", stream.ID))
```
pwilczynskiclearcode, line 143: This is really helpful... 
Maybe we could print out the links to `Google Cloud Run`, `Google big Table` with viewership metrics, and `Grafana` dashboard if they are easy to construct

PR #332, opened 2023-11-13
Check playback info and source playback

mjh1:
Check the playback info call without an API key works.
Check that source playback was available.

### DIFF
--- a/internal/app/recordtester/recordtester_app.go
+++ b/internal/app/recordtester/recordtester_app.go
@@ -349,7 +349,7 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 		}
 
 		if len(assets) != 1 {
-			return 247, fmt.Errorf("unexpected numnber of assets. expected: 1 actual: %d", len(assets))
+			return 247, fmt.Errorf("unexpected number of assets. expected: 1 actual: %d", len(assets))
 		}
 		if !assets[0].SourcePlaybackReady {
 			return 246, fmt.Errorf("source playback was not ready")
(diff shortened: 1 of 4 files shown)


internal/app/recordtester/recordtester_app.go
```diff
@@ -331,6 +335,33 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
...
+			Filters: map[string]interface{}{
+				"sourceSessionId": sess.ID,
+			},
+		})
+		if err != nil {
+			return 248, err
+		}
+
+		if len(assets) != 1 {
+			return 247, fmt.Errorf("unexpected numnber of assets. expected: 1 actual: %d", len(assets))
```
pwilczynskiclearcode, line 352: typo: `numnber` ;)

leszko (approved):

internal/app/recordtester/recordtester_app.go
```diff
@@ -301,8 +301,12 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 		return 0, err
 	}
 
+	lapiNoAPIKey := api.NewAPIClient(api.ClientOptions{
+		Server:      rt.lapi.GetServer(),
+		AccessToken: "", // test playback info call without API key
```
victorges, line 306: Do we not have an available API key to use here? We actually wanted to get rid of the API-key-less support, since it causes some billing issues for example. We have a log in there to eventually remove it, but the deprecation was paused. 

What's the idea here not to use an API key for this call?

PR #347, opened 2023-11-22
wrtc-lt: More improvements and features for load test

victorges:


cmd/webrtc-load-tester/roles/orchestrator.go
```diff
@@ -133,28 +135,39 @@ func initClients(cliFlags loadTestArguments) {
...
+	var stream *api.Stream
+	if args.StreamID != "" {
+		stream, err = studioApi.GetStream(args.StreamID, false)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve stream: %w", err)
+		}
+		glog.Infof("Retrieved stream with name: %s", stream.Name)
+	} else {
+		stream, err = studioApi.CreateStream(api.CreateStreamReq{
+			Name: "webrtc-load-test-" + time.Now().UTC().Format(time.RFC3339),
```
pwilczynskiclearcode, line 148: I'd either remove `webrtc-` prefix at all or make the prefix based on protocol (hls/webrtc)

pwilczynskiclearcode:
It would be good to expose it in github workflow

victorges:
> It would be good to expose it in github workflow

Just did! Instead of exposing the `stream-id` directly thou I added a custom "extra args" field, then we can customize any specific arg more easily

PR #355, opened 2024-04-03
tester: delete asset on completion

gioelecerati:


internal/app/recordtester/recordtester_app.go
```diff
@@ -363,10 +363,13 @@ func (rt *recordTester) Start(fileName string, testDuration, pauseDuration time.
 		if code, err := checkPlaybackInfo(assets[0].PlaybackID, rt.lapi, lapiNoAPIKey); err != nil {
 			return code, err
 		}
+
+		rt.lapi.DeleteAsset(assets[0].ID)
```
hjpotter92, line 367: shouldn't this be a `defer` call earlier, before one of the above `return` statements?

PR #357, opened 2024-06-19
record-tester: Further improvements for Catalyst E2E tests

victorges:
- Allow skipping source playback checks with a `--skip-source-playback` flag
- Make number of profiles check required, while specifying the default profiles if not provided
- Also refactored `record-tester` a bit to avoid copying around option vars.

If anyone is reviewing this, go commit by commit as the above are kept separate.


mjh1 (commented):
lgtm

